PR #27, opened 2019-05-10
Update getting-started-installation.md

tkarra:
Add a warning about how to secure deployments done with the tool ArangoDB Starter


Simran-B:
@tkarra Thanks a lot for your contribution to ArangoDB!

Simran-B:
Sorry for the delay and thanks for this addition!

PR #29, opened 2019-05-14
add information about withId parameter for PREGEL_RESULT function

jsteemann:
Updated docs for new parameter for PREGEL_RESULT AQL function

### DIFF
--- a/3.4/graphs-pregel.md
+++ b/3.4/graphs-pregel.md
@@ -69,7 +69,7 @@ db._createEdgeCollection("edges", {
 
 You will need to ensure that edge documents contain the proper values in their sharding attribute.
 For a vertex document with the following content `{ _key: "A", value: 0 }`
-the corresponding edge documents would have look like this:
+the corresponding edge documents would have to look like this:
 
 ```js
 { "_from":"vertices/A", "_to": "vertices/B", "vertex": "A" }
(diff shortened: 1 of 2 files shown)


3.4/graphs-pregel.md
```diff
@@ -154,9 +154,9 @@ The object returned by the `status` method might for example look something like
 ### Canceling an Execution / Discarding results
 
 To cancel an execution which is still running, and discard any intermediate results you can use the `cancel` method.
-This will immediately  free all memory taken up by the execution, and will make you lose all intermediary data. 
+This will immediately free all memory taken up by the execution, and will make you lose all intermediary data. 
 
-You might get inconsistent results if you cancel an execution while it is already in it's `done` state. The data is written
+You might get inconsistent results if you cancel an execution while it is already in its `done` state. The data is written
```
Simran-B: Seems like there is no way to tell if it's really done (including writing the results)?

3.5/graphs-pregel.md
```diff
@@ -173,22 +173,38 @@ AQL integration
...
+By default, the `PREGEL_RESULT` AQL function will return the `_key` of each vertex plus the result
+of the computation. In case the computation was done for vertices from different vertex collection,
+just the `_key` values may not be sufficient to tell vertices from different collections apart. In 
+this case, the `PREGEL_RESULT` can be given a second parameter `withId`, which will make it return
+the `_id` values of the vertices as well:
+
+```js
+FOR v IN PREGEL_RESULT(<handle>, true)
+  FILTER v.value >= 0.01
+  RETURN v._key
```
Simran-B: Shouldn't this return `v._id`?

3.4/graphs-pregel.md
jsteemann: I only know about `pregel.status(id)` but I don't know if that includes writing the results.

Simran-B: @graetzer is it possible to determine when it's really done (computation + writing)?

Simran-B: talked with Simon, he says it's possible to change it somehow so that you can differentiate between done processing and done writing (really done).

PR #30, opened 2019-05-15
Add Monitoring for DBServer cleanOutServer process 

OmarAyo:
ref: https://github.com/arangodb/planning/issues/3934

+ fixed some typos


3.4/administration-cluster.md
```diff
@@ -208,3 +208,53 @@ To distribute shards onto the new _DBServer_ either click on the
 `Distribute Shards` button at the bottom of the `Shards` page in every
 database.
 
+This process can be monitored using the following javascript, that can give a
```
Simran-B: `This` refers to the cleanOutServer process, but the previous paragraph talks about rebalancing shards. Not sure if we should move the new content up above the paragraph `After the _DBServer_ has been cleaned out...`  or if it would suffice to change `This` to `The cleanOutServer process`.

3.4/administration-cluster.md
```diff
@@ -208,3 +208,53 @@ To distribute shards onto the new _DBServer_ either click on the
...
+
+The output should be similar to the one below:
+
+```
+arangosh --server.username root --server.password pass --javascript.execute ~./serverCleanMonitor.js -- DBServer0002
+[7836] INFO Server DBServer0002 count is 1
+
+arangosh --server.username root --server.password pass --javascript.execute ~./serverCleanMonitor.js -- DBServer0002
+[7842] INFO Server DBServer0002 count is 0
+```
```
Simran-B: The user is supposed to invoke arangosh from time to time and check if the count reached 0 already. Would it be beneficial if the script would instead poll the server every so often and sleep in between, until the count reaches zero, and print a status message every now and then so that you know it still does something?

3.4/administration-cluster.md
OmarAyo: I guess that it makes sense to rename it as you said:

```
The cleanOutServer process
```

Will change it now

3.4/administration-cluster.md
OmarAyo: Script was update by this [commit](https://github.com/arangodb/docs/pull/30/commits/d6538448d09b55432447cca564a55d0db86710bc)

3.4/administration-cluster.md
```diff
@@ -208,3 +208,57 @@ To distribute shards onto the new _DBServer_ either click on the
 `Distribute Shards` button at the bottom of the `Shards` page in every
 database.
 
+The clean out process can be monitored using the following script,
+which periodically prints the amount of shards that still need to be moved.
+It is basically a countdown to when the process finishes.
+The current status is logged every 5 seconds. You may adjust it by
+changing the number on the line `internal.wait(5);`.
```
Simran-B: Note to self: change this sentence here and update the script (sleep duration as second parameter).

PR #38, opened 2019-05-22
Removed AQL

cw00dw0rd:
The above mentioned guarantees are not currently offered in AQL only graph module. (confirmed with @mchacki)


3.4/graphs.md
```diff
@@ -56,7 +56,7 @@ However the graph module adds an additional layer on top of these collections gi
 * If you delete a vertex all edges will be deleted, you will never have loose ends
```
Simran-B: Maybe mention here that you *can* have loose ends if you start with a dataset which has dangling edges to begin with?

cw00dw0rd:
```suggestion
* If you delete a vertex all edges will be deleted
```
Removing the last part and then adding something in the following paragraph about if the dataset starts with dangling edges they will not be removed upon creation seems like a good idea.

PR #53, opened 2019-05-29
Improve sidebar hiding

Simran-B:
On mobile devices (<800px) we close the vertical navigation. We don't want it to play the closing animation (CSS transition).


Simran-B:
Having it flicker isn't nice either, but I can't think of another solution at the moment.

PR #69, opened 2019-06-05
add pregel `shardKeyAttribute` to documentations

ObiWahn:
https://github.com/arangodb/team-orco-planning/issues/135


ObiWahn:
https://github.com/arangodb/team-orco-planning/issues/135

Simran-B:
The default value is `vertex`, what does that actually mean? Does it expect a top-level attribute called "vertex" on the edge?

PR #71, opened 2019-06-11
Prototype for version-less chapters (kube-arangodb) + overview page

Simran-B:
Awaiting management decision


Simran-B:
Top-level version switcher will remain for the foreseeable future. It would be difficult to get rid of it unless version-specific content would be moved into a single page each, so that each page could have its own selector.

Content duplication was reduced by symlinking some of the content: #534

PR #79, opened 2019-06-18
Netlify: Bypass lack of symlinks by copying, make redirects dynamic

Simran-B:
**WIP**

Attempt to read the versions from _config.yml

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -1,1 +1,1 @@
-/              /docs/3.4/
+/              /docs/devel/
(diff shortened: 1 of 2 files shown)


Simran-B:
Seems like something is run async because failure is logged before some shell commands in the script.

Simran-B:
#487 might fix this as a side effect.

Simran-B:
See #585. Can't make redirects dynamic as long as we symlink the stable and devel versions (same redirect files means same redirect target, unless we do JavaScript shenanigans).

PR #80, opened 2019-06-18
Bug fix/rename analyzer stem norm properties

Dronplane:
Renamed text and norm analyzer parameters.
Fixed default value for accent in norm analyzer section.


Simran-B:
In what version were the attributes changed in code? One of the RCs or before?

Dronplane:
Just changed, should be available in next RC

PR #88, opened 2019-06-24
update security options

ObiWahn:


Simran-B:
@ObiWahn Is this ready for merging?

PR #89, opened 2019-06-25
ArangoSearch refactor

Simran-B:
DocuBlocks: https://github.com/arangodb/arangodb/pull/9597


Simran-B:
HTTP docs:
- there is no attribute `"properties": { ... }` expected, but rather the View properties directly (same level as `name` and `type`). This needs to be fixed (pointed out by Vadim).
- there are multiple places where the response object is described, but without `code` and `error`
- `result` is not mentioned / described as containing the important query result
- `globallyUniqueId` is not mentioned and missing a description - it's an internal attribute as per Vadim, which collections have as well.

PR #107, opened 2019-07-18
Add 2 missing ways to call COLLECT.

ObiWahn:


Simran-B:
Only backported to 3.3 and 3.4 because older versions are EoL

PR #112, opened 2019-07-22
Clarification: installation via `kubectl` works for default namespace only

Simran-B:
Ported from https://github.com/arangodb/arangodb/pull/8412 by jgerken

Note that we have yet to decide whether the docs repo should contain the documentation for the operator or if it should stay here:

https://github.com/arangodb/kube-arangodb/blob/master/docs/Manual/Deployment/Kubernetes/Usage.md


Simran-B:
Possibly related:
https://github.com/arangodb/arangodb/issues/10951

PR #113, opened 2019-07-22
Prepare docs for 3.6

Simran-B:
Copied 3.5 to 3.6, because Oskar script constructs target folder from CMakeLists.txt, which is 3.6 in devel.


Simran-B:
Need to add symlinks `_data/3.6-program-options-arango*.json` once the examples are committed in `generated`.

PR #126, opened 2019-07-30
add options to disable the foxx store / api

ObiWahn:


ObiWahn:
@Simran-B one of the logs suggests that i should be adding something to `_layouts/default.html`.
There are only the following files:

```
» find -iname "*default*"
./.git/logs/refs/remotes/origin/kubectrl-nondefault-namespace
./.git/refs/remotes/origin/kubectrl-nondefault-namespace
./_layouts/default.html
```

the `_layouts/default.html` there does not reference the security options in 3.5 or devel. Therefor I am not what needs to be done to fix this PR.

jsteemann:
I think the new page still needs to be linked in the manual.

3.4/security-security-options.md
```diff
@@ -0,0 +1,29 @@
+---
+layout: default
+description: arangod provides a variety of options to make a setup more secure
+---
+# Server security options
+
+_arangod_ provides a variety of options to make a setup more secure.
```
Simran-B: "variety" for 2 options seems a little exaggerated. We shouldn't this paragraph as-is from 3.5 with a whole lot more options.

ObiWahn:
Thank you @Simran-B  I was just going to fix the problem!

PR #131, opened 2019-08-05
Version placeholder in macOS commands

Simran-B:


OmarAyo:
I think now is clearer and should avoid any future misunderstanding - thanks simran

PR #137, opened 2019-08-09
Add cluster limitations

graetzer:


Simran-B:
I think we should change the Transactions chapter in the Manual as well to cover "old-style" and streaming transactions, similar to how it is already in the HTTP book. Then this remark could be added to the manual rather then the HTTP docs...

graetzer:
feel free to do so, the exact limitations still need to be agreed upon

PR #139, opened 2019-08-17
Update examples for 3.3

Simran-B:
Added remark: `--rocksdb.enforce-block-cache-size-limit` option disabled (not supported because of bugs in RocksDB 5.6, 3.4.x ships with 5.16+)


Simran-B:
@sleto-it Added the warning box:
https://deploy-preview-139--zealous-morse-14392b.netlify.com/docs/3.3/administration-configuration-rocks-db.html#caching

PR #146, opened 2019-08-21
Updated 'compute' with 'concatenate'

divyanshidm:


Simran-B:
While you are right that the example **concatenates** document attributes and an inline string, the point of the example is to show that you can use expressions to _compute_ new values. Maybe my word choice isn't the best. Should it be _create_ instead?

A different example:

    FOR user IN users RETURN POW(user.value, 2) // value squared

Simran-B:
Made it a bit clearer:
https://github.com/arangodb/docs/commit/55c266a37c7a71f91929776b8673ea255e1832ad
https://github.com/arangodb/docs/commit/a9b8c44ffb5b177eab5f1ec6119d1428b575015e

PR #148, opened 2019-08-21
Typo fixes

sonyarianto:


3.5/release-notes-new-features35.md
```diff
@@ -86,7 +86,7 @@ Execution plan:
   5   ReturnNode             1       - RETURN doc
 ```
 
-Execution plan with a the primary sort order of the index being utilized:
+Execution plan with a primary sort order of the index being utilized:
```
Simran-B: There is only a single sort order, so I think we should keep the `the` over the `a`:
```suggestion
Execution plan with the primary sort order of the index being utilized:
```

PR #149, opened 2019-08-21
Deprecation warning for DATE_COMPARE()

Simran-B:
`DATE_COMPARE()` compares datestring substring and can thus not utilize any indexes.

Since we added `DATE_TRUNC()` there is no use for it anymore and therefore I would like to get it removed in v4.0.0.

In Addition, I think we should revise `DATE_ADD()` / `DATE_SUBSTRACT` and `DATE_DIFF()` and possibly change their behavior in v4.0.0 to make them more useful / behave as expected / return correct results. Should warnings be added for that as well, or will it be sufficient to note in the upgrading-changes-40.md in the future?

### DIFF
--- a/3.6/aql/functions-date.md
+++ b/3.6/aql/functions-date.md
@@ -579,6 +579,11 @@ with decimal places.
 
 Check if two partial dates match.
 
+{% hint 'warning' %}
+This function is deprecated and will be removed in v4.0.0.
+You may use `DATE_TRUNC(date1, unit1) == DATE_TRUNC(date2, unit2)` instead.
+{% endhint %}
+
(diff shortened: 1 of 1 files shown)


jsteemann:
IMHO for DATE_ADD|SUBTRACT we should add another parameter that toggles the behavior change.
That way migration from 3.x to 4.0 will be easy, even though if we change the default value of that parameter in 4.0.

Simran-B:
Discussed with Jörg that this is a 4.0 topic (date handling, possibly with timezone support and custom internal type/representation) and that it is sufficient to deprecate things in 3.8 if we plan to remove something in 4.0.

Simran-B:
Closing this PR for now, suggested deprecation is in API 4.0 Google doc.

PR #151, opened 2019-08-22
Some minor fixes

sonyarianto:


Simran-B:
Thanks again! Do you want to add more commits to this PR later perhaps?

sonyarianto:
@Simran-B Thank you, maybe later for other commits, now only that :)

sonyarianto:
@Simran-B I just want ArangoDB become better and better :) Because I love the product :)

PR #159, opened 2019-08-27
Add TLS v1.3 as SSL protocol

Simran-B:
https://github.com/arangodb/arangodb/pull/9817
https://github.com/arangodb/arangodb/pull/9818

To be updated in other places as well, e.g. driver documentation.


Simran-B:
Couldn't find anything regarding the encryption protocol in driver docs, seems to be limited to server-side.

PR #160, opened 2019-08-27
Update Known Issues and add 3.6 pages

jsteemann:


Simran-B:
To be clarified:
- https://github.com/arangodb/release-qa/issues/50
- https://github.com/arangodb/planning/issues/3756
- Upgrading ArangoSearch Views from 3.4.x to 3.5.0 pre-releases is not supported
- Calling a shutdown endpoint may not result in a proper shutdown while the node/server is still under load.
- The web interface shows a wrong replication mode in the replication tab in Active Failover deployments sometimes.

KVS85:
> To be clarified:
> 
> * Upgrading from 3.4.4 to 3.4.5 under CentOS fails - are more recent versions affected as well?
>   [arangodb/release-qa#50](https://github.com/arangodb/release-qa/issues/50)
Fixed in 3.4.8 and 3.5.0. Will work only for subsequent updates (i.e. 3.4.8 -> 3.4.9, 3.4.8 -> 3.5.0 -> 3.5.1).

> * Shutdown of leader instance hangs, customer says it didn't happen again?
>   [arangodb/planning#3756](https://github.com/arangodb/planning/issues/3756)
Don't know.

> * Upgrading ArangoSearch Views from 3.4.x to 3.5.0 pre-releases is not supported
Fixed in RC4+. Need 3.4.7 to have good upgrade.

> * Calling a shutdown endpoint may not result in a proper shutdown while the node/server is still under load.
Don't know.

> * The web interface shows a wrong replication mode in the replication tab in Active Failover deployments sometimes.
Wasn't fixed directly.

> * Cluster + AQL + wrong custom shard key (uncommon case) - accidentally removed from 3.5 known issues?
>   [arangodb/arangodb#6399](https://github.com/arangodb/arangodb/issues/6399)
Don't know.

PR #161, opened 2019-08-27
SmartGraphs examples not rendered correctly

Simran-B:
image

The blocks were apparently not migrated properly.


Simran-B:
Blocked on #9833, can't get the 3rd example to work in 3.5.0

Simran-B:
Discovered bug (server-side?) which prevents dropping graph collections. Team Gordo is on it.

PR #165, opened 2019-09-06
Added missing keywords

ifcologne:


ifcologne:
Btw. should we add the pseudo-variables `OLD` and `NEW` to the variables section as well?

3.5/aql/fundamentals-syntax.md
```diff
@@ -121,18 +121,22 @@ The complete list of keywords is currently:
   <li>INBOUND</li>
   <li>INSERT</li>
   <li>INTO</li>
+  <li>K_SHORTEST_PATHS</li>
   <li>LET</li>
   <li>LIMIT</li>
   <li>NONE</li>
   <li>NOT</li>
   <li>NULL</li>
+  <li>OPTIONS</li>
```
jsteemann: Technically this is not true.
Neither options `OPTIONS` nor `SEARCH` nor `TO` are keywords.
They are used by the query language, yes, but they aren't keywords.

Keywords you can't use as attribute names without quoting, e.g.
```
arangod> db._query(`RETURN { foo: "bar", k_shortest_paths: 123 }.k_shortest_paths`);
JavaScript exception in file '/home/jsteemann/Arango35NoAsan/js/server/modules/@arangodb/arango-statement.js' at 84,16: ArangoError 1501: AQL: syntax error, unexpected K_SHORTEST_PATHS keyword, expecting identifier or quoted string or bind parameter or [ near 'k_shortest_paths: 123 }.k_shorte...' at position 1:22 (while parsing)
```
But `OPTIONS`, `SEARCH` and `TO` aren't keywords, so it's ok to use them as attribute names:
```
arangod> db._query(`RETURN { foo: "bar", options: 123 }.options`);
[object GeneralArrayCursor, count: 1, cached: false]
[ 
  123 
]

arangod> db._query(`RETURN { foo: "bar", search: 123 }.search`);
[object GeneralArrayCursor, count: 1, cached: false]
[ 
  123 
]

arangod> db._query(`RETURN { foo: "bar", to: 123 }.to`);
[object GeneralArrayCursor, count: 1, cached: false]
[ 
  123 
]
```

Simran-B: From further up the document:

> [...] keywords can only be used at certain locations in a query.
> This also makes all keywords reserved words that must not be used for other purposes than they are intended for.
> 
> For example, it is not possible to use a keyword as a collection or attribute name.
> If a collection or attribute need to have the same name as a keyword, the collection or attribute name needs to be quoted.

What are `OPTIONS`, `SEARCH` etc. then if not keywords? Or are they keywords, but not reserved words?

jsteemann: They are not keywords, as shown above: you can use these names anywhere in queries you like, e.g. as attribute names, variable names etc.
They are just regular names. The intention here is to not use so many keywords, because each keyword we add will make certain user queries (that use the keyword as an attribute or variable name) and cause support effort.
So when the grammar allows it, we try to get away without new keywords.

For example, let's look at how `SEARCH` is handled (which isn't a keyword):
Example query: `FOR doc IN view SEARCH ...`
When the parser is directly before the `SEARCH`, it is at the end of parsing a regular `FOR` statement. The only tokens that are allowed to follow here are top-level statements, e.g.: `RETURN`, `COLLECT`, `LET`, `FILTER`, INSERT`, `UPDATE`, `REPLACE`, `REMOVE`, `UPSERT` etc. 
But definitely no unquoted string must follow here, e.g. `FOR doc IN view meow`. So, what we did is that we indeed allow a `FOR` declaration to be followed by an unquoted string, but then require it to be `SEARCH`. If the following unquoted string is not `SEARCH`, we make things fail.
The same is true for `TO`, which seems to be only used for `SHORTEST_PATH` style queries. Here it is easy to assert that the unquoted string between the two vertices must be `TO`, and make everything else just fail.

I don't know a good name for this, but in the end these things aren't keywords. They are language words, but without the full restrictions that keywords imply.

Simran-B: How about adding a second list with the non-keywords?

```markdown
The following additional words are used in language constructs, but are not keywords / reserved words and thus do not need to be quoted if used as attribute names etc.:

- K_SHORTEST_PATHS
- OPTIONS
- SEARCH
- SHORTEST_PATH
- TO
```

Simran-B: Is LIKE a keyword?

Simran-B: SHORTEST_PATH, K_SHORTEST_PATHS and LIKE are actually keywords.
OPTIONS, SEARCH and TO are not keywords.

Simran-B:
I went through the list of (key-)words using a simple query `FOR doc IN coll RETURN doc.<word>` and wrote down which are keywords (and their subtype), which are keyword-like and which are not keywords, see https://github.com/PrismJS/prism/pull/2047

Simran-B:
An unquoted string literal is called an *identifier*:
https://github.com/arangodb/arangodb/blob/devel/arangod/Aql/grammar.y#L247

3.5/aql/operators.md
```diff
@@ -329,21 +345,30 @@ AQL provides array operators `[*]` for
...
+- `=` variable assignment (LET operation)
+- `WITH` with operator (WITH, UPDATE, REPLACE, COLLECT operations)
+- `INTO` into operator (INSERT, UPDATE, REPLACE, REMOVE, COLLECT operations)
 - `||` logical or
 - `&&` logical and
-- `**`, `!=` equality and inequality
-- `IN` in operator
+- `OUTBOUND`, `INBOUND`, `ANY`, `ALL`, `NONE` graph traversal direction and array comparison modifiers
+- `==`, `!=`, `LIKE`, `=~`, `!~` equality, inequality, wildcard string match, regex match, regex non-match
+- `IN`, `NOT` in operator, logical negation
```
jsteemann: Should be just `IN, NOT IN`. Logical negation should be removed here.

PR #125, opened 2019-07-30
HotBackup documentation for 3.5

neunhoef:
Related PRs:
- https://github.com/arangodb/arangodb/pull/10056
- https://github.com/arangodb/arangodb/pull/9934
- https://github.com/arangodb/arangodb/pull/10055
- https://github.com/arangodb/arangodb/pull/9932

PR #183, opened 2019-10-08
Added Hotbackup limitations.

joerg84:


Simran-B:
Please see this commit:
https://github.com/arangodb/docs/pull/125/commits/907dd7de509525268507bcac04eba4bdd28e4405

As per Vadim, DC2DC support should be called "experimental".

PR #189, opened 2019-10-21
Remove NoResultsNode

goedderz:


3.6/aql/execution-and-performance-optimizer.md
```diff
@@ -352,9 +352,8 @@ The following execution node types will appear in the output of `explain`:
   attribute). Will appear exactly once in a query that contains an *UPDATE* statement.
 * *UpsertNode*: upserts documents in a collection (given in its *collection*
   attribute). Will appear exactly once in a query that contains an *UPSERT* statement.
-* *NoResultsNode*: will be inserted if *FILTER* statements turn out to be never
-  satisfiable. The *NoResultsNode* will pass an empty result set into the processing
-  pipeline.
+* *NoResultsNode*: _Unused since 3.6_. The *NoResultsNode* will pass an empty
```
Simran-B: If it's removed in 3.6, can't we just delete the description? It should obviously be mentioned in the CHANGELOG of the main repo however (TBD)

goedderz: It's still in there, we need it for backwards compatibility during rolling upgrades, that's why I left it. But you're right and its probably best to just remove it from the documentation.

Simran-B: I see. Will remove it then.

PR #196, opened 2019-11-02
update linux os configuration instructions

jsteemann:


3.6/installation-linux-osconfiguration.md
```diff
@@ -130,6 +103,55 @@ To make the settings durable, it will be necessary to store the adjusted
 settings in /etc/sysctl.conf or other places that the operating system is
 looking at.
 
+Memory Limits
+-------------
+
+A long-running arangod process may use a lot of virtual memory after a while,
```
joerg84: minor s/use/accumulate

jsteemann: Applied that suggestion.

PR #206, opened 2019-11-06
Documentation/add arangosearch known issue (652, part 2)

KVS85:
https://github.com/arangodb/backlog/issues/652


KVS85:
> @KVS85 Added arangosh example as well as missing table header.
> 
> Should we replace the example by something that creates the _analyzers collections for all DBs automatically?

I don't think so, because noone was hurt.

PR #208, opened 2019-11-07
Adjust Gemfile.lock to require newer nokogiri

Simran-B:
jekyll-algolia plugin is not maintained, hence no deps bump to be expected


Simran-B:
Resolves #195

PR #219, opened 2019-11-11
add documentation for query timeout

ObiWahn:
I think https://www.arangodb.com/docs/stable/aql/invocation-with-arangosh.html#setting-options is not the best location for options. I think It is a bit too hidden.

### DIFF
--- a/3.6/release-notes-new-features36.md
+++ b/3.6/release-notes-new-features36.md
@@ -132,7 +132,10 @@ into a set of equal-distance buckets.
 
 ### Miscellaneous AQL changes
 
-ArangoDB 3.6 provides a new AQL function `GEO_AREA` for area calculations.
+ArangoDB 3.6 provides the following new AQL functionality:
+
+- function `GEO_AREA` for area calculations
+- query option `timeout` to restrict the execution time.
 
 HTTP API extensions
 -------------------
--- a/3.6/release-notes-new-features36.md
+++ b/3.6/release-notes-new-features36.md
@@ -134,8 +134,8 @@ into a set of equal-distance buckets.
 
 ArangoDB 3.6 provides the following new AQL functionality:
 
-- function `GEO_AREA` for area calculations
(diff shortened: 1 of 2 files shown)


3.6/release-notes-new-features36.md
```diff
@@ -132,7 +132,10 @@ into a set of equal-distance buckets.
 
 ### Miscellaneous AQL changes
 
-ArangoDB 3.6 provides a new AQL function `GEO_AREA` for area calculations.
+ArangoDB 3.6 provides the following new AQL functionality:
+
+- function `GEO_AREA` for area calculations
```
```suggestion
- a function `GEO_AREA()` for area calculations
```

3.6/release-notes-new-features36.md
```diff
@@ -132,7 +132,10 @@ into a set of equal-distance buckets.
...
+- query option `timeout` to restrict the execution time.
```
```suggestion
- a query option `timeout` to restrict the execution time.
```

ObiWahn:
> I agree. How else can the query option be set? In the query itself, like `OPTIONS { timeout: 123 }`?
No. It can be only set via the options parameter of `_query`. Only some operations like `INSERT` have extra options in AQL.
> Does it expect a number in seconds?
yes

Simran-B:
will be backported to 3.5.4: https://github.com/arangodb/arangodb/pull/10405/files

Simran-B:
> Only some operations like `INSERT` have extra options in AQL.

The FOR operation supports OPTIONS for index hints, but RETURN does not support any OPTIONS. Maybe a good feature request?

jsteemann:
Things such as a query timeout shouldn't be specified in per-operation `OPTIONS`, as there can be multiple operations in the same query.
For example
```
FOR doc IN view OPTIONS { timeout: 10 }
  UPDATE doc._key WITH { ... } IN collection OPTIONS { timeout: 5 }
  RETURN NEW OPTIONS { timeout: 30 }
```
So it will really be better to set the timeout outside of per-operation `OPTIONS`.

PR #224, opened 2019-11-14
Release Notes: Incompatible Changes 3.6 (--vst.maxsize)

jsteemann:


3.6/release-notes-upgrading-changes36.md
```diff
@@ -30,3 +30,113 @@ above will make the function return `null` and trigger a warning too. Example:
 ```js
 DATE_SUBTRACT("2018-08-22T10:49:00+02:00", 100000, "years") // null
 ```
+
```
Simran-B: Are these changes really supposed to be in `3.6/release-notes-upgrading-changes36.md`, which has the headline "Incompatible changes in ArangoDB 3.6"?

jsteemann: Yes, because it is an incompatible change.
3.5 and before did not restrict the date ranges explicitly. So there was no guarantee what date ranges worked. 
But 3.5 may have worked for some dates before `0000-00-00` or values after `9999-12-31`. That's why the restrictions in 3.6 make it backwards-incompatible to 3.5.

Simran-B: Sorry, I meant the changes below: various additions to the startup options as well as the extension of the `/_api/database*` endpoints. Were things also removed or only added? In the latter case I don't see how that would be a breaking change?!

jsteemann: True, these are no breaking changes.

jsteemann: I will adjust the PR then

jsteemann: Done, please check!

PR #235, opened 2019-11-21
Update 3.6 release docs for ArangoSearch

gnusi:
Internal ref:
- https://docs.google.com/document/d/1NrwrTfLx15Y0aaDyAkUQPLsQth_baCNBz6OCHTLF_X8/edit?usp=sharing


3.6/release-notes-new-features36.md
```diff
@@ -12,12 +12,128 @@ here.
...
+
+```js
+FOR d IN documentSource // documentSource can be either collection or ArangoSearch View
+SORT d.foo
+LIMIT 100
+RETURN d
+```
+
+For the collection case the optimization is possible IFF:
+- there is an index of type `hash`, `skip` or `edge` picked by the optimizer
```
jsteemann:
```suggestion
- there is an index of type `hash`, `skiplist` or `edge` picked by the optimizer
```
Does this also work with indexes of type "persistent"? At least for the RocksDB storage engine, the "persistent" index is identical to hash & skiplist indexes in terms of functionality, so it should also be supported here.

3.6/release-notes-new-features36.md
```diff
@@ -12,12 +12,128 @@ here.
...
+- `endMarker` : `<string>`, default: ""<br>
+  this value will be appended to ngrams at the beginning of input sequence
+
+- `streamType` : `"binary"|"utf8"`, default: "binary"<br>
+  type of the input stream
+
+### Added "Edge ngram" support to `text` analyzer
+
+The following optional properties can be provided with `text`
+Analyzer definition:
```
jsteemann:
```suggestion
analyzer definition:
```

Simran-B: I consider Analyzers a feature name and capitalized the first letter in most of the documentation, so this should stay as-is.

3.6/release-notes-new-features36.md
Simran-B: Also what about the `primary` index?

iurii-i-popov: 1. The optimization works for rocksdb only. So "persistent" is excluded.
2. Primary index is on:
`FOR d IN collectionName FILTER d._key IN ['c0', 'c1'] SORT NOOPT(d._key) DESC LIMIT 1 RETURN d`.

jsteemann: RocksDB has a persistent index. Why is it excluded?

3.6/release-notes-new-features36.md
jsteemann: Then it should be capitalized consistently. I think it isn't right now.

3.6/release-notes-new-features36.md
iurii-i-popov: Sorry, it is not hasCovering(). Forget about rocksdb.

Simran-B: > The optimization works for rocksdb only.

No late materialization if you use MMFiles storage engine, makes sense.

> So "persistent" is excluded. / it is not hasCovering().

I'm confused about this. In RocksDB, the index types `hash`, `skiplist` and `persistent` are identical and we even dropped the first two in the Web UI. Does it work with all three index type? In MMFiles, I don't expect any index type including `persistent` to work of course.

3.6/release-notes-new-features36.md
Simran-B: Should be fixed by https://github.com/arangodb/docs/pull/235/commits/12f6cd12c04d17c6fee4bcd494c73089739028b2

The remaining occurrences should refer to a variable:

```
const analyzers = await db.listAnalyzers();
// analyzers is an array of Analyzer descriptions
```

3.6/release-notes-new-features36.md
iurii-i-popov: @Simran-B You are right. The persistent index is included also. Thank you!

PR #236, opened 2019-11-22
Remove GATHER and REMOTE from the subquery splicing naughty list

markuspf:
GATHER and REMOTE do not prevent subquery splicing anymore.


markuspf:
This probably needs to be merged with #233

PR #243, opened 2019-11-25
Add documentation about options used in database creation

ObiWahn:


3.6/data-modeling-databases-working-with.md
```diff
@@ -138,8 +138,13 @@ change into the current database to the new database. Changing the current
 database must explicitly be requested by using the
 *db._useDatabase* method.
 
-The *options* attribute currently has no meaning and is reserved for
-future use.
+The *options* attribute can be used to set defaults for collections that will
+be created in the new database:
+
+* *sharding*: The sharding method to use. Valid values are: "" or "single
```
jsteemann:
```suggestion
* *sharding*: The sharding method to use. Valid values are: "" or "single".
```

3.6/data-modeling-databases-working-with.md
```diff
@@ -138,8 +138,13 @@ change into the current database to the new database. Changing the current
...
+* *replicationFactor*: Default replication factor. Special values include "satellite",
+  which will replicate the collection to every DB-server, and 1, which disables replication.
```
jsteemann:
```suggestion
  which will replicate the collection to every DB-server, and "1", which disables replication.
```

ObiWahn: It has to be a number or a special string value. Strings will not be converted to numbers.

ObiWahn:
The descriptions of options should really be in a single place. Is there some plan to remove the split of the documentation?

3.6/data-modeling-databases-working-with.md
jsteemann: I understand. "replicationFactor" can be a string "satellite" or a numeric.
But the value for "minReplicationFactor" should always be numeric.
The change I suggest above is not about turning the `1` into a string, but to make the value `1` stand a bit more out from the rest of the text formatting-wise. We could also put it inside asterisks or backticks if you prefer, but it should stand out a bit from the rest of the text.

jsteemann:
I agree. @Simran-B : are there any changes planned for cleaning up the documentation "mess", i.e. removing the need to adjust two repositories when docs are adjusted?

3.6/data-modeling-databases-working-with.md
```diff
@@ -141,10 +141,13 @@ database must explicitly be requested by using the
...
 be created in the new database:
 
-* *sharding*: The sharding method to use. Valid values are: "" or "single
+* *sharding*: The sharding method to use. Valid values are: "" or "single".
+  Setting this option to "single" will enable the one-shard feature in the enterprise edition.
 * *replicationFactor*: Default replication factor. Special values include "satellite",
   which will replicate the collection to every DB-server, and 1, which disables replication.
-* *minReplicationFactor*: Default minimum replication factor.
+* *minReplicationFactor*: Default minimum replication factor. Collections with
+  a lower replication factor than this value can not be created in the new
```
jsteemann: This is actually _not_ what this attribute is doing. 
Can you check other occurrences of "minReplicationFactor" in the docs or changelog or new features to see what it does exactly?

Simran-B:
> are there any changes planned for cleaning up the documentation "mess", i.e. removing the need to adjust two repositories when docs are adjusted?

@jsteemann Do you refer to the DocuBlock being in the main repo and the rest being here? Well, I discussed this with Jörg last week and we do want to put everything into the docs repo. This isn't an easy task however, and we may do additional refactoring as part of a larger project soon.

3.6/data-modeling-databases-working-with.md
```diff
@@ -138,8 +138,16 @@ change into the current database to the new database. Changing the current
...
+The *options* attribute can be used to set defaults for collections that will
+be created in the new database:
+
+* *sharding*: The sharding method to use. Valid values are: "" or "single".
+  Setting this option to "single" will enable the one-shard feature in the enterprise edition.
+* *replicationFactor*: Default replication factor. Special values include "satellite",
+  which will replicate the collection to every DB-server, and 1, which disables replication.
+* *minReplicationFactor*: Default minimum replication factor. If there are
+  less than minReplication replicas available the collection will become
+  read-only.
```
Simran-B: This sounds a lot like the description of the write-concern option. Do both exist? write-concern as startup option but minReplicationFactor in the API? This question was also raised by Simon:
https://docs.google.com/document/d/1wvLKGnBPXTXvDY-N_blgyOhSudMCaUrP6sa6cHKcQ3U/edit?usp=sharing

If this means the same thing then it should have the same name, WDYT?

PR #252, opened 2019-12-02
added docu block hook for metrics

kvahed:
https://github.com/arangodb/arangodb/pull/10423


jsteemann:
Please also add the metrics feature to the new features document and please add the new startup option for disabling the metrics API to the manual.

PR #255, opened 2019-12-04
Default value of commands is useless / removed in 3.6.devel

Simran-B:
Internal ref:
- https://github.com/arangodb/planning/issues/3306

arangodb/arangodb#10620 removes default attribute for commands, so we should check for the presence of this attribute


Simran-B:
image

To be tested with the default attribute not present in JSON data

PR #260, opened 2019-12-06
Add two missing example settings for unattended installation under Linux

Simran-B:
https://github.com/arangodb/arangodb/pull/9286

### DIFF
--- a/3.6/installation-linux.md
+++ b/3.6/installation-linux.md
@@ -42,6 +42,8 @@ For unattended installations, you can set the password using the
 ```
 echo arangodb3 arangodb3/password password NEWPASSWORD | debconf-set-selections
 echo arangodb3 arangodb3/password_again password NEWPASSWORD | debconf-set-selections
+echo arangodb3 arangodb3/upgrade string no | debconf-set-selections
+echo arangodb3 arangodb3/storage_engine string 1 | debconf-set-selections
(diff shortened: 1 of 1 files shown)


Simran-B:
Thanks a lot for your contribution to ArangoDB @black0range!

Simran-B:
From /debian/3.5/common/postinst (oskar):

```bash
db_get @EDITION@/storage_engine
STORAGE_ENGINE=$RET
...
# fill in correct storage engine into arangod.conf
sed -i /etc/arangodb3/arangod.conf -e "s;storage-engine = auto;storage-engine = $STORAGE_ENGINE;"
```

Looks to me like it should be possible to set the storage engine by name instead of number?

    echo arangodb3 arangodb3/storage_engine string rocksdb | debconf-set-selections

Is it really `no`?

    echo arangodb3 arangodb3/upgrade string no | debconf-set-selections

```bash
# check if we should upgrade the database directory
UPGRADE=false #requires upgrade
...
db_get @EDITION@/upgrade #wants upgrade
if [ "$RET" = "true" ];  then
    if [ "$UPGRADE" = "true" ];  then
```

I also wonder if it should be optional (commented out).

ansoboleva:
3.6 has reached EOL. Obsolete PR.

PR #269, opened 2019-12-12
Release notes 3.6: fix a few query plan issues

jsteemann:


3.6/release-notes-new-features36.md
```diff
@@ -363,20 +355,22 @@ Execution plan:
...
  13   IndexNode         DBS   1000000     - FOR doc IN test   /* primary index scan, index only, projections: `_key`, 5 shard(s) */
   4   UpdateNode        DBS         0       - UPDATE doc WITH #3 IN test
   7   RemoteNode        COOR        0       - REMOTE
-  8   GatherNode        COOR        0       - GATHER
+  8   GatherNode        COOR        0       - GATHER   /* parallel */
 ```
 
-The optimization will also work with filters:
+As can be seen above, the benefit of applying the optimization is that the extra
+communication between the coordinator and database server is removed. This will
```
```suggestion
communication between the Coordinator and the DB-Server is removed. This will
```

3.6/release-notes-new-features36.md
```diff
@@ -363,20 +355,22 @@ Execution plan:
...
+mean less cluster-internal traffic and thus can result in faster execution. Thus the
```
```suggestion
mean less cluster-internal traffic and thus can result in faster execution. The
```

PR #281, opened 2020-01-05
Update getting-started-installation.md

RobertShaw1:


Simran-B:
Cellar is homebrew specific if I'm not mistaken, so macOS. But I think the description assumes Linux as operating system. The path probably varies not only based on OS, but also installation method. Not sure what to do. We could either link to the specific installation instructions or add an overview to the getting started page to cover most OSes and installation methods.

fceller:
If everything installed correctly, there should be a symlink

```
fceller@FranksMacBook ~/1/o/w/dmg> ls -l /usr/local/sbin/arangod
lrwxr-xr-x  1 fceller  admin  37  6 Jan 13:34 /usr/local/sbin/arangod -> ../Cellar/arangodb/3.5.3/sbin/arangod
```

You can use

`brew link --overwrite arangodb`

to regenerate the links.

PR #285, opened 2020-01-07
Known issues v3.6.0

Simran-B:
https://docs.google.com/document/d/1ha6X1JwZgEv6kDzhEhcyeyXwCdQD9qRX3ptNnQS6lLI/edit#

- [x] macOS DMG package not notarized
  __added KI to other versions as well__

- [x] Creation of collection in OneShard database doesn’t honor maximum possible replication factor
  __removed__

- [x] rebalanceShards API works in _system database only
  __removed__

- [x] UI is unstable (possible layout problems)
  __removed__

- [x] Windows installer failed to detect whether we need to Upgrade
  __was already added in December for 3.5 and 3.6. Should it be added to 3.4 as well? Is only the Enterprise Edition installer affected?__

- [x] Windows installer: %PATH% isn’t populated with installed binaries path
  __added for 3.5 and 3.6. Should it be added to 3.4 as well?__

- [x] failed to locate javascript.startup-directory
  __is only the client package / arangosh affected? Could this be related to the RC package (maybe not using the correct paths to locate .conf files?)__

- [x] In progress jobs are not reset/restarted in case of a Foxxmaster failover
  __added, but unsure if the description is accurate__

- [x] Hypergraph functionality broken (edges pointing to other edges)
  __added__


KVS85:
> * [ ]  Windows installer failed to detect whether we need to Upgrade
>   **was already added in December for 3.5 and 3.6. Should it be added to 3.4 as well? Is only the Enterprise Edition installer affected?**

Yes. 3.4 also has this. Both Community and Enterprise are affected. But, this problem is only for patch releases. I.e. 3.4 to 3.5 and 3.5 to 3.6 are OK. Whenever 3.4.A to 3.4.B aren't.

> * [ ]  Windows installer: %PATH% isn’t populated with installed binaries path
>   **added for 3.5 and 3.6. Should it be added to 3.4 as well?**

Yes.

> * [ ]  failed to locate javascript.startup-directory
>   **is only the client package / arangosh affected? Could this be related to the RC package (maybe not using the correct paths to locate .conf files?)**

It presents in GA as well. The problem is that in `%INSTALLDIR%\etc\arangodb3\arangosh.conf` there is the following section:
```
[javascript]
startup-directory = @ROOTDIR@/share/arangodb3/js
```
While JS is in `@ROOTDIR@/share/arangodb3`. That is improper packaging and, probably, presents in 3.5 and 3.4 as well.

PR #294, opened 2020-01-13
Parallel Move Shard

maierlars:


3.7/release-notes-new-features37.md
```diff
@@ -8,3 +8,10 @@ Features and Improvements in ArangoDB 3.7
 The following list shows in detail which features have been added or improved in
 ArangoDB 3.7. ArangoDB 3.7 also contains several bug fixes that are not listed
 here.
+
+
+Cluster
+-------
+
+### Parallel Move Shard
+Shards can now move in parallel. The old locking mechanism was replaced by a read-write-lock and thus allows multiple jobs for the same _to_ server. The actual transfere rates are still limited on dbserver side but there is a huge overall speedup. This also affects `CleanOutServer` and `ResignLeadership` jobs.
```
jsteemann:
```suggestion
Shards can now move in parallel. The old locking mechanism was replaced by a read-write-lock and thus allows multiple jobs for the same _to_ server. The actual transfer rates are still limited on dbserver side but there is a huge overall speedup. This also affects `CleanOutServer` and `ResignLeadership` jobs. 
```

Simran-B:
@maierlars I changed `to server` into `destination server`, or is that incorrect?

maierlars:
> @maierlars I changed `to server` into `destination server`, or is that incorrect?
I approve that change.

maierlars:
image

PR #297, opened 2020-01-13
added missing description of "global" query parameter in replication

Simran-B:
https://github.com/arangodb/arangodb/pull/10840

Shows a different text as description of global option on WAL page?!

https://deploy-preview-297--zealous-morse-14392b.netlify.com/docs/3.6/http/replications-walaccess.html#tail-recent-server-operations


Simran-B:
https://www.arangodb.com/docs/stable/http/replications-walaccess.html#tail-recent-server-operations
Need to check what is going on here.

PR #306, opened 2020-01-20
Start with 3.7 release notes (new features etc.)

jsteemann:
Depends on https://github.com/arangodb/arangodb/pull/10930


3.6/release-notes-new-features36.md
```diff
@@ -212,23 +212,16 @@ read only documents that are absolutely necessary to compute the query result,
...
-`SORT`+`LIMIT` combination, e.g:
-
-```js
-FOR d IN documentSource // documentSource can be either a collection or an ArangoSearch View
-  SORT d.foo
-  LIMIT 100
-  RETURN d
-```
+In 3.6 the optimization can only be applied to queries retrieving data from a
+collection or an ArangoSearch and that contain a SORT`+`LIMIT` combination, e.g:
```
```suggestion
collection or an ArangoSearch View and that contain a `SORT`+`LIMIT` combination, e.g:
```

3.6/release-notes-new-features36.md
```diff
@@ -241,6 +234,13 @@ For the ArangoSearch View case the optimization is possible if and only if:
 - all attribute accesses can be covered by stored attributes
   (e.g. using `primarySort`)
 
+```js
+FOR d IN myView
+  SORT d.foo
+  LIMIT 100
```
```suggestion
  LIMIT 100  // only 100 documents will be materialized
```

3.6/release-notes-new-features36.md
```diff
@@ -241,6 +234,13 @@ For the ArangoSearch View case the optimization is possible if and only if:
 - all attribute accesses can be covered by stored attributes
```
Simran-B: The View needs to "own" the data, right? As in, the data on disk and any regular collection indexes are not usable for ArangoSearch, correct?
```suggestion
- all attribute accesses can be covered by attributes stored in the View index
```

3.6/release-notes-new-features36.md
```diff
@@ -241,6 +234,13 @@ For the ArangoSearch View case the optimization is possible if and only if:
...
   (e.g. using `primarySort`)
 
+```js
+FOR d IN myView
+  SORT d.foo
```
Simran-B: What is sorted on here needs to match (a prefix of) the primarySort configuration to be able to materialize late, right?
```suggestion
  SORT d.foo // if the View has "primarySort": [ { "field": "foo", "direction": "asc" } … ]
```

In a later example we say:

```js
// Given we have primary sort on fields ["foo", "bar"]
FOR d IN myView
  SEARCH d.foo == "baz"
  SORT d.bar DESC    // field "bar" will be read from the View
```
Does the direction of the first field matter?
- `"primarySort": [ { "field": "foo", "direction": "asc" }, { "field": "bar", "direction": "desc" } … ]`
- `"primarySort": [ { "field": "foo", "direction": "desc" }, { "field": "bar", "direction": "desc" }  … ]`

gnusi: Late materialization is applicable IFF `LIMIT` node is there:
```
FOR d IN myView
  SEARCH d.foo == 'bar'
  SORT d.bar DESC
  LIMIT 100
  RETURN d
```
order of fields in `primarySort` doesn't matter

Please note that this particular optimization is for 3.7, 3.6 doesn't have it. In 3.6 we can do late materialization for queries like
```
FOR d IN myView
  SEARCH d.foo == 'bar'
  SORT BM25(d) DESC // no reference to document attributes
  LIMIT 100
  RETURN d
```
only

Simran-B: Summary of Slack discussion:
- no late document materialization (LDM) in 3.6 if an attribute is referenced in SORT operation, i.e. the following query can benefit from sort optimization, but no LDM can be applied:
  ```
  FOR d IN myView
    SORT d.foo
    LIMIT 100
    RETURN d
  ```
- but there is LDM in 3.7 for this case. `primarySort` leads to ArangoSearch indexing the attribute values (has its own copy of the data), so that this index can be consulted. The field order and sorting direction in primarySort does not matter for LDM, but only for the sorting optimization!
- In 3.7, users can also instruct ArangoSearch to store more attribute values (like `"storeValues": "value"`), so that a View query may be fully covered without bothering the storage engine

Simran-B: Addendum: There is LDM in 3.6 for:
```
FOR d IN myView
  SORT d.foo
  LIMIT 100
  RETURN d
```
… if `primarySort` contains `foo`, but the `direction` is `desc` (`"asc": false`), i.e. the sort optimization can **not** be applied but the LDM can instead. There is **no need for LDM** if we can read straight from the sorted index (sort optimization).

PR #308, opened 2020-01-21
Added recommendation not to use NFS.

joerg84:
Know issues with NFS are performance and hard links required for Hot Backup.

### DIFF
--- a/3.5/installation-linux-osconfiguration.md
+++ b/3.5/installation-linux-osconfiguration.md
@@ -16,19 +16,20 @@ File Systems
 ------------
 
 We recommend to **not** use BTRFS on linux, as user have reported issues using it in
-conjunction with ArangoDB. We experienced ArangoDB facing latency issues when accessing 
-its database files on BTRFS partitions. In conjunction with BTRFS and AUFS we also saw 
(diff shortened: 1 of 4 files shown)


3.5/installation-linux-osconfiguration.md
```diff
@@ -16,19 +17,22 @@ File Systems
 ------------
 
 We recommend to **not** use BTRFS on linux, as user have reported issues using it in
-conjunction with ArangoDB. We experienced ArangoDB facing latency issues when accessing 
-its database files on BTRFS partitions. In conjunction with BTRFS and AUFS we also saw 
+conjunction with ArangoDB. We experienced ArangoDB facing latency issues when accessing
+its database files on BTRFS partitions. In conjunction with BTRFS and AUFS we also saw
 data loss on restart.
 
+We would not recommend to use NFS for performance reasons, furthermore we
```
dothebart: I would rather put it like: 
We would not recommend network filesystems such as NFS ... since all this is true for AFS, SMB, ...

Simran-B: Updated: https://github.com/arangodb/docs/commit/b488da5a31e368c68c9fa7629a6634420778c523

PR #309, opened 2020-01-21
Add audit log events for hotbackup to documentation.

neunhoef:
3.6 and 3.7


Simran-B:
Also [3.5](https://github.com/arangodb/arangodb/pull/10958), which is not merged as of now, but merging the documentation change anyway.

PR #310, opened 2020-01-21
Start HTTP/2 docs

graetzer:


graetzer:
TODO:
- entry in release notes
- page highlighting the pro / cons of using http2
- maybe a blogpost

Issue #317, opened 2020-01-27
Render new program startup option fields `os` and `component`


Simran-B:
https://github.com/arangodb/arangodb/pull/10974

```
"os": [
  "linux",
  "macos",
  "windows"
],

"component": [
  "single",
  "agent",
  "coordinator",
  "dbserver"
]
```

Simran-B:
Tracked by DOC-104

PR #323, opened 2020-01-31
Render new startup option info about os / component

Simran-B:
Ref: #317


Simran-B:
https://deploy-preview-323--zealous-morse-14392b.netlify.com/docs/3.7/programs-arangod-options.html#cluster

The logic to hide the information if all OSes/components accept an option is a bit brittle because it checks the array length, but the amount of elements is unlikely to change.

PR #325, opened 2020-02-03
Feature/add array replace nth function

dothebart:
documentation change for https://github.com/arangodb/arangodb/pull/11042


3.7/aql/functions-array.md
```diff
@@ -546,7 +546,8 @@ Replace the element at *position* in *anyArray* with *replaceValue*.
 
 - **anyArray** (array): array with elements of arbitrary type
 - **position** (number): the position of the element to replace. Positions start
-  at 0. Negative positions are not supported.
+  at 0. Negative positions are supported, with -1 being the last array element.
+  If a negative *position* is out of bounds, the array is returned unmodified.
```
Simran-B: @dothebart  Shouldn't the behavior be symmetric with out-of-bounds positive positions? Raise an error if the index points before the first element (or two before it?) unless a padding value is provided leading to `[newValue, pad1, pad2, padN, ...oldValues]`?

dothebart: make this an if < 0 throw? 
```
    replaceOffset = (static_cast<int64_t>(length) + posParam < 0) ? 0: static_cast<uint64_t>(length +  posParam);
```

Simran-B: It would be pretty uncommon. Since we raise a runtime error when no padding value is provided and the position > length, I think it would make sense to throw also if a negative position points below the first element (or two before 0 to behave analogous to positive positions, where one element after the last is allowed?).

PR #329, opened 2020-02-03
Add --server.allow-deprecated-storage-engine

Simran-B:
Not sure if the option is supposed to be documented or if it's only for internal use (tests):
https://github.com/arangodb/arangodb/pull/10994

Anyway, we should tell users whether they can still use MMFiles in the upgrade case or not (MMFiles code removal postponed to 4.0?)


jsteemann:
@simran-b: unclear yet, let's leave this open as a reminder that it still needs clarifying.

ifcologne:
Still open. We are currently waiting for feedback from the community. 
We announced it during the 3.7 roadmap webinar and the blog post on MMFiles deprecation (afaik not published yet).

Simran-B:
Closing, as MMFiles will be fully removed in 3.7.0

PR #345, opened 2020-02-25
Add edgeCollections and vertexCollections traversal options.

danielhlarkin:
Document the recently added `vertexCollections` and `edgeCollections` restrictions for traversal options.


Simran-B:
@danielhlarkin 

I created anonymous graph `A/1 --[x/?]--> B/2 --[y/?]--> C/3` to see what happens if you restrict the edge collections using the new option **in addition** to specifying the edge collection set.

```js
FOR v,e,p IN 0..10 OUTBOUND "A/1" x, y
OPTIONS { edgeCollections: ["x"], vertexCollections: ["A","B","C"] }
RETURN v._key
```

Looks like the option can be used to further restrict the traversal, which is fine. Just something we might want to document.

Something I noticed in the query explain:

```
Traversals on graphs:
 Id  Depth  Vertex collections  Edge collections  Options                                                            Filter / Prune Conditions
 2   0..10                      x                 uniqueVertices: none, uniqueEdges: path, vertexCollections: A,B,C
```

There are columns for vertex collections and edge collections, but only the edge collection "x" is listed the latter. The vertex collections are listed in the options column instead. Shouldn't it be like this?

```
Traversals on graphs:
 Id  Depth  Vertex collections  Edge collections  Options                                  Filter / Prune Conditions
 2   0..10  A,B,C               x                 uniqueVertices: none, uniqueEdges: path
```

PR #354, opened 2020-03-04
Update information about version

vasekd:


Simran-B:
Hi @vasekd, thank you for your contribution to ArangoDB!

PR #357, opened 2020-03-12
Satellite graphs docs draft

markuspf:
First draft of Satellite Graph Docs for the Alpha release.


3.7/graphs-satellite-graphs.md
```diff
@@ -0,0 +1,134 @@
...
+collections to graphs. All of the usual benefits and caveats apply.
+
+When doing joins involving graph traversals, shortest paths, or k-shortest paths
+in an ArangoDB cluster, data has to be exchanged between different servers. In particular
+graph traversals are usually executed on a Coordinator, because they need global information. 
+
+This results in a lot of network traffic and slow query execution.
+
+Satellite graphs are synchronously replicated to all DB-Servers that are part of
+a cluster, which enables DB-Servers to execute parts of joins locally.
```
goedderz:
```suggestion
a cluster, which enables DB-Servers to execute graph traversals (and (k-)shortest paths), and possibly joins with traversals, locally.
```

3.7/graphs-satellite-graphs.md
```diff
@@ -0,0 +1,134 @@
...
+
+This greatly improves performance for such joins.
```
goedderz:
```suggestion
This greatly improves performance for such queries.
```

3.7/graphs-satellite-graphs.md
```diff
@@ -0,0 +1,134 @@
...
+To create a satellite graph in arangosh, use the `satelliteGraph` module.
+
+    arangosh> satGM = require("@arangodb/satellite-graph")
+    arangosh> satGM._create("satelliteGraph", [ satGM._relation("edges", "vertices", "vertices") ], [], {});
+
+You can also create vertex and edge collections manually, and set their `replicationFactor` to
+`satellite`
+
+    arangosh> db._createDocumentCollection("vertices2", { replicationFactor: "satellite" });
+    arangosh> db._createEdgeCollection("edges2", { replicationFactor: "satellite" });
```
goedderz: That alone does not suffice; `distributeShardsLike` has to be set to the leading collection of the graph, which is the first vertex collection added to it.

3.7/graphs-satellite-graphs.md
```diff
@@ -0,0 +1,134 @@
...
+a cluster, which enables DB-Servers to execute parts of joins locally.
+
+This greatly improves performance for such joins.
+
+To create a satellite graph in arangosh, use the `satelliteGraph` module.
+
+    arangosh> satGM = require("@arangodb/satellite-graph")
+    arangosh> satGM._create("satelliteGraph", [ satGM._relation("edges", "vertices", "vertices") ], [], {});
+
+You can also create vertex and edge collections manually, and set their `replicationFactor` to
```
hkernbach: I think I would remove that section completely, as the user needs also to add _distributeShardsLike to the collection marked as initial collection inside the graph definition.

3.7/graphs-satellite-graphs.md
goedderz: Also, a line how these would be added to the actual graph would be good.

3.7/graphs-satellite-graphs.md
```diff
@@ -0,0 +1,134 @@
...
+
+    arangosh> var explain = require("@arangodb/aql/explainer").explain
+    arangosh> graphModule = require("@arangodb/general-graph");
+    arangosh> satGraphModule = require("@arangodb/satellite-graph");
+    arangosh> graphModule._create("normalGraph", [ graphModule._relation("edges", "vertices", "vertices") ], [], {});
+    arangosh> satGraphModule._create("satelliteGraph", [ satGM._relation("satEdges", "satVertices", "satVertices") ], [], {});
+    arangosh> db._create("collection", {numberOfShards: 8})
+    arangosh> db._create("collection2", {numberOfShards: 8})
+
+Let's analyse a join involving a traversal.
```
hkernbach:
```suggestion
Let's analyse a join involving a traversal:
```

3.7/graphs-satellite-graphs.md
```diff
@@ -0,0 +1,134 @@
...
+Caveats
+-------
+
+The cluster will automatically distribute all satellite graphs to all DBServers, and apply
+synchronous replication. 
+This means that writes are executed on the leader only, and then have to be replicated
+to all followers, leading to degradation in write performance.
+
+There is also a chance that a write has not been replicated in time before a query is executed
+and a query involving satellite graphs returns out of date results.
```
goedderz: If writes are done while a query is being instantiated, there is chance that a write has not been fully replicated. This can lead to inconsistent results.
```suggestion
If writes are done to a satellite graph while a query is being instantiated, there is chance that a write has not been fully replicated. This can lead to inconsistent results for queries involving that graph.
```

3.7/graphs-satellite-graphs.md
```diff
@@ -0,0 +1,134 @@
...
+
+Traversals on graphs:
+ Id  Depth  Vertex collections  Edge collections  Options                                  Filter / Prune Conditions
+ 3   1..1   vertices            edges             uniqueVertices: none, uniqueEdges: path                           
+
+Optimization rules applied:
+ Id   RuleName
+  1   scatter-in-cluster
+  2   remove-unnecessary-remote-scatter
+```
```
goedderz: I think it's possible to specify the commands and let the output be generated with the documentation. But its been a long time since I've used that last, so it might have changed. @Simran-B, can you shed some light on that?

3.7/graphs-satellite-graphs.md
markuspf: Whats the difference between "instantiation" and "execution"?

I neither like "inconsistent" (because it doesn't say what it's inconsistent *with*), nor "out of date".

goedderz: > Whats the difference between "instantiation" and "execution"?

Technically, after the planning stage, the query snippets get sent to all DBServers. There, they are instantiated, meaning an object is created in the query registry. Most importantly, the local transactions are opened, and thus RocksDB Snapshots are taken on each DBServer.

Only if writes happen while the transactions are started, it can happen that one DBServer gets a write before the transaction is opened, and another does not.

> I neither like "inconsistent" (because it doesn't say what it's inconsistent with), nor "out of date".

What I meant with inconsistent, and I agree that we should clarify that, is that the different replicas of the same satellite collection(s) have a different state (snapshot) at the very beginning of the query. Thus the same traversal will return different results on different DBServers. That can lead to all kind of breakage, depending on the query. We could even say this is undefined behaviour, as of course all optimizer rules assume this does not happen.

PR #360, opened 2020-03-13
add description for LIKE arangosearch function

gnusi:


Simran-B:
Escaping doesn't seem to work, Andrey will investigate.

PR #367, opened 2020-03-17
Document memory override env variable.

neunhoef:


neunhoef:
Please note that this is something which is implemented for 3.6.3 and 3.7. We might want to wait with the publication until 3.6.3 is out. On the other hand, the text explains that the option is only available for >= 3.6.3, so there is no actual problem.

PR #370, opened 2020-03-18
 Encryption recommendation.

joerg84:


joerg84:
3.6 only, needs devel after checking.

PR #371, opened 2020-03-18
Tell users about unsupported remote schemas in main documentation.

ObiWahn:


Simran-B:
Note that I already added a hint box further down:
image

PR #374, opened 2020-03-18
Snowball 2: Update supported languages

Simran-B:
I added a hint regarding lexicographical sorting, but I think it's quite right.

`FOR doc IN coll SORT doc.letter RETURN doc` sorts according to `--default-language`, but so does 
`FOR doc IN view SORT doc.letter RETURN doc`.

What seems to be affected are range queries, e.g.
`FILTER doc.letter >= "X" AND doc.letter <= "Þ" SORT doc.letter`. This returns the expected range of characters for Icelandic using a collection, but not using a View:
`SEARCH ANALYZER(doc.letter >= "X" AND doc.letter <= "Þ", "text_is") SORT doc.letter`

Andrey confirmed that SEARCH does not honor the Analyzer locale nor the defined server language. It's unclear whether the behavior is undefined or comparing characters codepoints.

We should probably add this to the Known Issues list!

Also see https://github.com/arangodb/arangodb/pull/11046#issuecomment-581773274


3.7/arangosearch-analyzers.md
```diff
@@ -367,3 +364,50 @@ Name       | Type       | Language
...
+language-dependent tokenization and normalization. The ICU data file
+`icudtl.dat` that ArangoDB ships with contains information for a lot of
+languages, which are technically all supported.
+
+{% hint 'warning' %}
+The alphabetical order of characters is not taken into account by ArangoSearch,
+i.e. range queries and sorting of Views will not follow the language rules as
+per the defined Analyzer locale or the server startup option
+`--default-language`!
+{% endhint %}
```
gnusi: I think that'd be better to attach this notice to the description of supported SEARCH functions.
`IN_RANGE, >, >=, <, <=, STARTS_WITH` are affected

PR #376, opened 2020-03-19
Hot Backups: Restoration required maintenance mode in Active Failover setups

Simran-B:
This is a workaround to avoid non-restored followers becoming leader.

### DIFF
--- a/3.5/programs-arangobackup-limitations.md
+++ b/3.5/programs-arangobackup-limitations.md
@@ -48,9 +48,12 @@ When restoring hot backups in Active Failover setups, it is necessary to
 prevent that a non-restored follower becomes leader by temporarily setting
 the maintenance mode:
 
-1. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"on"'`
+1. `curl -X PUT <endpoint>/_admin/cluster/maintenance -d'"on"'`
 2. Restore the Hot Backup
-3. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"off"'`
+3. `curl -X PUT <endpoint>/_admin/cluster/maintenance -d'"off"'`
+
+Substitute `<endpoint>` with the actual endpoint of the **leader**
+single server instance.
 
 Identical Minor Version
 -----------------------
--- a/3.5/programs-arangobackup-limitations.md
+++ b/3.5/programs-arangobackup-limitations.md
@@ -42,6 +42,16 @@ dropping of cluster databases, collections and indexes.
 It must be ensured that for the hot backup no such changes are made to the
 cluster's inventory, as this could lead to inconsistent hot backups.
 
+### Active Failover Special Limitations
+
+When restoring hot backups in Active Failover setups, it is necessary to
+prevent that a non-restored follower becomes leader by temporarily setting
+the maintenance mode:
+
+1. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"on"'`
+2. Restore the Hot Backup
+3. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"off"'`
+
 Identical Minor Version
 -----------------------
 
@@ -152,4 +162,3 @@ Hot backups are not automatically replicated between instances. This is
 true for both the Active Failover setup with 2 (or more) single servers
 and for the Datacenter to Datacenter Replication between clusters.
 Simply take hot backups on all instances.
-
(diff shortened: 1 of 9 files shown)


3.5/programs-arangobackup-limitations.md
```diff
@@ -42,6 +42,16 @@ dropping of cluster databases, collections and indexes.
 It must be ensured that for the hot backup no such changes are made to the
 cluster's inventory, as this could lead to inconsistent hot backups.
 
+### Active Failover Special Limitations
+
+When restoring hot backups in Active Failover setups, it is necessary to
+prevent that a non-restored follower becomes leader by temporarily setting
+the maintenance mode:
+
+1. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"on"'`
```
KVS85: Is that resilensingle endpoint?

3.5/programs-arangobackup-limitations.md
```diff
@@ -42,6 +42,16 @@ dropping of cluster databases, collections and indexes.
...
+2. Restore the Hot Backup
+3. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"off"'`
+
```
KVS85: Is that resilensingle endpoint?

3.6/programs-arangobackup-limitations.md
```diff
@@ -37,6 +37,16 @@ dropping of cluster databases, collections and indexes.
 It must be ensured that for the hot backup no such changes are made to the
 cluster's inventory, as this could lead to inconsistent hot backups.
 
+### Active Failover Special Limitations
+
+When restoring hot backups in Active Failover setups, it is necessary to
+prevent that a non-restored follower becomes leader by temporarily setting
+the maintenance mode:
+
+1. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"on"'`
```
KVS85: Is that resilensingle endpoint?

3.6/programs-arangobackup-limitations.md
```diff
@@ -37,6 +37,16 @@ dropping of cluster databases, collections and indexes.
...
+2. Restore the Hot Backup
+3. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"off"'`
```
KVS85: Is that resilensingle endpoint?

3.7/programs-arangobackup-limitations.md
```diff
@@ -37,6 +37,16 @@ dropping of cluster databases, collections and indexes.
 It must be ensured that for the hot backup no such changes are made to the
 cluster's inventory, as this could lead to inconsistent hot backups.
 
+### Active Failover Special Limitations
+
+When restoring hot backups in Active Failover setups, it is necessary to
+prevent that a non-restored follower becomes leader by temporarily setting
+the maintenance mode:
+
+1. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"on"'`
```
KVS85: Is that resilensingle endpoint?

3.7/programs-arangobackup-limitations.md
```diff
@@ -37,6 +37,16 @@ dropping of cluster databases, collections and indexes.
...
+2. Restore the Hot Backup
+3. `curl -X PUT <leader-endpoint>/_admin/cluster/maintenance -d'"off"'`
```
KVS85: Is that resilensingle endpoint?

3.5/programs-arangobackup-limitations.md
Simran-B: @KVS85 Not sure what you mean. You should talk to the leader instance, not a follower. Is the leading instance's endpoint called "resilentsingle"? It's the first time I hear about it.

KVS85: I meant there are agency and single server (resilentsingle) for each leader/follower node. Is that a call to single server endpoint?

Simran-B: So the question is whether the leading agent or the leading single server is supposed to be contacted. Given the path I think it is the leading single server, the agency paths are different aren't they?

maxkernbach: `/_admin/cluster/maintenance` is the endpoint of the **leader DBserver** (leading single server)

It is also possible to enable/disable the maintenance mode via the agency, but the DBserver approach is the easier one. So all that has to be done is rename `<leader-endpoint>` into `<leader-single-endpoint>` or similar.

PR #391, opened 2020-03-30
ArangoJS: Node.js support

Simran-B:
Remove mentions of concrete versions.

@pluma Is the statement about IE11 still correct?


Simran-B:
Also, the examples features v4.0. What should we replace it with? Latest?

```
# for version 4.x.x
yarn add arangojs@4
# - or -
npm install --save arangojs@4
```

pluma:
> Also, the example features v4.0. What should we replace it with? Latest?
> 
> ```
> # for version 4.x.x
> yarn add arangojs@4
> # - or -
> npm install --save arangojs@4
> ```

I think that should be 6 now. The point is to show how to install a specific (older) version instead of the latest release.

Simran-B:
Thanks, adjusted.

PR #393, opened 2020-03-30
Add remark about required Analyzer features for PHRASE() & NGRAM_MATCH()

Dronplane:
Added important note for analyzers to be used with NGRAM_MATCH  and PHRASE functions


Simran-B:
@Dronplane Does it return incorrect results if the Analyzer doesn't have the features enabled, or does it raise an error?

Dronplane:
It just won't find anything

PR #397, opened 2020-04-01
java-reference-graalvm-native

rashtao:


3.7/drivers/java-reference-graalvm-native.md
```diff
@@ -0,0 +1,13 @@
+---
+layout: default
+description: GraalVM Native Image
+---
+# GraalVM Native Image
+
+The library supports GraalVM Native Image generation since version `6.6.1`. The related configuration can be found
```
joerg84, line 7: s/library/driver?

rashtao: fixed --> driver

rashtao:
> Wouldn't this information be more suited for this page? https://www.arangodb.com/docs/stable/drivers/java-getting-started.html

As you prefer! Do you mean also with a nested menu entry under `Getting Started`?
Feel free to adjust it as you think is best...

PR #415, opened 2020-04-16
Driver connection pooling

Simran-B:
Looks like the Go driver has a setting ConnLimit. Is that all it takes for connection pooling?


informalict:
> Looks like the Go driver has a setting ConnLimit. Is that all it takes for connection pooling?

Yes it has and java driver has it too, but a user can create:
- one poll of connections with e.g. ConnLimit = 100000 or
- many polls of connections. E.g. 100 polls and each of that poll has ConnLimit = 1000

in both situations ConnLimit is 1000000 and of course it is too big

PR #422, opened 2020-05-04
Fix typo: SERACH -> SEARCH

ex-nerd:


Simran-B:
Well spotted and thanks for reporting, added a fix for all affected versions in #423

PR #424, opened 2020-05-05
added docs for "optimize-count" optimizer rule

jsteemann:
Docs PR for https://github.com/arangodb/arangodb/pull/11542

### DIFF
--- a/3.7/release-notes-new-features37.md
+++ b/3.7/release-notes-new-features37.md
@@ -230,7 +230,8 @@ collection scan or an index scan, without any additional filtering on document
 attributes (early pruning or document post-filtering) and without using LIMIT.
 
 The optimization will help in the following situation:
-```
+
+```js
 FOR doc IN collection
   LET count = COUNT(
     FOR sub IN subCollection
@@ -247,8 +248,9 @@ no LIMIT clause and no FILTER condition or calculation that requires
 accessing document data. Accessing index data is supported for filtering (as
 in the above example that would use the edge index), but not for further 
 calculations.
+
 In case a subquery does not match these criteria, it will not use the 
-optimized code path for counting, but will executed normally.
+optimized code path for counting, but will execute normally.
 
 If the optimization is triggered, it will show up in the query execution
 plan under the rule name `optimize-count`.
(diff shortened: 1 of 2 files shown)


3.7/release-notes-new-features37.md
```diff
@@ -221,6 +221,38 @@ benefit from this change straight away. This also holds true for subqueries,
...
+### Count optimizations
+
+Subqueries can now use an optimized code path for counting documents if they
+are supposed to only return the number of matching documents.
+The optimization will be triggered for read-only subqueries that use a full 
+collection scan or an index scan, without any additional filtering on document
+attributes (early pruning or document post-filtering) and without using LIMIT.
+
+The optimization will help in the following situation:
+```
```
Simran-B, line 233:
```suggestion

```js
```

3.7/release-notes-new-features37.md
```diff
@@ -221,6 +221,38 @@ benefit from this change straight away. This also holds true for subqueries,
...
+
+The restrictions are that the subquery result must only be used with the
+`COUNT`/`LENGTH` AQL function and not for anything else. The subquery itself 
+must be read-only (no data-modification subquery), not use nested FOR loops,
+no LIMIT clause and no FILTER condition or calculation that requires
+accessing document data. Accessing index data is supported for filtering (as
+in the above example that would use the edge index), but not for further 
+calculations.
+In case a subquery does not match these criteria, it will not use the 
+optimized code path for counting, but will executed normally.
```
Simran-B, line 251:
```suggestion
optimized code path for counting, but will execute normally.
```

3.7/release-notes-new-features37.md
```diff
@@ -221,6 +221,38 @@ benefit from this change straight away. This also holds true for subqueries,
...
+```
+
+The restrictions are that the subquery result must only be used with the
+`COUNT`/`LENGTH` AQL function and not for anything else. The subquery itself 
+must be read-only (no data-modification subquery), not use nested FOR loops,
+no LIMIT clause and no FILTER condition or calculation that requires
+accessing document data. Accessing index data is supported for filtering (as
+in the above example that would use the edge index), but not for further 
+calculations.
+In case a subquery does not match these criteria, it will not use the
```
Simran-B, line 250:
```suggestion

In case a subquery does not match these criteria, it will not use the 
```

PR #442, opened 2020-05-22
ES-554 known issue

KVS85:
Known issue of ES-54 with only in 3.7+ fix (no backports will happen).

The following workaround for previous (3.6-) versions is present, but I don't know how to reference or where to add it (Confluence isn't an option):
```
We need to force DBServer purge analyzer cache.
To do so we will need intermediate analyzer (just dummy, we will remove it later)

The steps are the following (in arangosh connected to Coordinator):

   1. var a = require("@arangodb/analyzers"); and a.remove("<PROBLEMATIC ANALYZER>");
   2. a.save("dummy_analyzer_<SOME SUFFIX TO HAVE NON-EXISTING ANALYZER>", "identity", {});
   3. db._query("FOR d IN  <ANY EXISTING VIEW NAME> SEARCH ANALYZER(STARTS_WITH(d.test, 'something'), '<ANALYZER FROM STEP 2>') RETURN d"); ← this query  with new analyzer will force cache purging. If this query will report error about missing analyzer, than  wait a minute and repeat until it succeeds. 
   4. Create analyzer <PROBLEMATIC ANALYZER> with new, correct properties.
   5. a.remove("<ANALYZER FROM STEP 2>");
   6. Check faulting query. There is possibility that first checks will still report analyzer missing. But in a minute this should pass and the query will execute normally. DB server cache should be in order this time.
```


Simran-B:
@KVS85 Why isn't this an option? https://arangodb.atlassian.net/wiki/spaces/DEVSUP/overview
For tech alerts we created pages on the website, would that work? https://www.arangodb.com/alerts/
We can inline the workaround into the known issue page, but it will require some HTML because Markdown tables don't allow new lines for Markdown lists

KVS85:
> @KVS85 Why isn't this an option? https://arangodb.atlassian.net/wiki/spaces/DEVSUP/overview
Confluence have problems to export anonymous pages. We tried to settle it but weren't pleased.

> For tech alerts we created pages on the website, would that work https://www.arangodb.com/alerts/
AFAIK that is for real issues that are harmful for data and *are* fixed in all supported versions  their presence.

> We can inline the workaround into the known issue page, but it will require some HTML because Markdown tables don't allow new lines for Markdown lists
I believe that it could be a separate workarounds page(s) with link(s) from KI.

Simran-B:
As there doesn't seem to be any other scenario in which one would want to purge the Analyzer cache, I'm adding the workaround description below the ArangoSearch KI list (3.5, 3.6).

PR #443, opened 2020-05-24
Added Env variables to API change notes.

joerg84:


joerg84:
IMO Env variables are also part of the API and should be mentioned here.

PR #453, opened 2020-06-02
Fix external links on mobile

Simran-B:
In Chrome for Android, clicking one of the reference links does not open a new tab with the absolute URL, but just blanks the main area.


Simran-B:
Seems like a blank return in click() handler using jQuery is interpreted like `return false` in Chrome on Android, but not on desktop. Returning false is the equivalent of 

                    event.preventDefault();
                    event.stopPropagation();

which is the opposite of what we want in this situation

https://stackoverflow.com/questions/11184276/return-false-from-jquery-click-event

PR #487, opened 2020-07-02
Relative redirects

Simran-B:
Allows us to avoid version-specific redirects

Symlink files and folders for Oasis, drivers and release notes / upgrading.
TODO: images need to be moved and links updated, but only after [Oasis doc PR](#466) is merged (it renames images).


Simran-B:
Superseded by #534

PR #499, opened 2020-07-14
Spring converter

rashtao:
Updated the Spring converter, since it was based on a deprecated class.


Simran-B:
Reverted changes to 3.4, 3.5, 3.6, 3.8. These versions will be symlinked by #487

PR #500, opened 2020-07-14
Java Driver: added documentation about ArangoCursor.close

rashtao:


Simran-B:
Reverted changes to 3.4, 3.5, 3.6, 3.8. These versions will later be symlinked by #487

PR #501, opened 2020-07-14
Spring data indexing documentation

rashtao:
added indexing limitations to spring data documentation


Simran-B:
Reverted changes to 3.4, 3.5, 3.6, 3.8. These versions will later be symlinked by #487

PR #508, opened 2020-07-22
Known issue: Request hangs when destroying AQL Streaming Cursor in Cluster

Simran-B:
https://arangodb.atlassian.net/browse/BTS-126

Unclear what other versions might be affected, please clarify.


rashtao:
`3.5.5` and `3.7.1-rc1` are not affected

Simran-B:
Observed in 3.7.0 however, added it to the 3.7 known issues.

PR #509, opened 2020-07-22
Applicability limitations of create document with overwrite parameter

rashtao:


Simran-B:
Contains monkey-patched `3.7/generated/allComments.txt`, to be merged after https://jenkins01.arangodb.biz/view/Documentation/job/arangodb-ANY-examples/659/

PR #521, opened 2020-08-10
Add list of langs accepted by our ICU

Simran-B:
Action plan:
- Gather test cases from ICU, Snowball and maybe other vendors for the most important languages
- Maybe involve the community to help testing languages
- List the most important langs in the docs, not only for ArangoSearch (`SORT`, `--default-language`, Fulltext index?)


Simran-B:
https://github.com/unicode-org/icu/blob/master/icu4c/source/test/testdata/WordBreakTest.txt

PR #535, opened 2020-08-28
Added a clarification on intermediate commits and upsert

mchacki:
Before I get this out of Draft I need to back port the statement to all released versions.

### DIFF
--- a/3.3/aql/operations-upsert.md
+++ b/3.3/aql/operations-upsert.md
@@ -86,7 +86,7 @@ explicitly.
 To make sure data are durable when an update query returns, there is the *waitForSync* 
 query option.
 
-In order to not accidentially update documents that have been written and updated since 
+In order to not accidentally update documents that have been written and updated since 
 you last fetched them you can use the option *ignoreRevs* to either let ArangoDB compare 
 the `_rev` value and only succeed if they still match, or let ArangoDB ignore them (default):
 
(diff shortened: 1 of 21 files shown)


jsteemann (approved):

mchacki:
Use the same approved comment in all versions 3.4 -> devel

Simran-B:
From Jörg's [Upsert notebook](https://github.com/joerg84/ArangoDBUniversity/blob/master/Upsert.ipynb):

> ## UPSERT Performance and Caveats
>
> The lookup attribute(s) from the search expression should be indexed in order to improve UPSERT performance Ideally, the search expression contains the shard key, as this allows the lookup to be restricted to a single shard
> 
> The lookup and the insert/update/replace parts are executed non- atomically. That means if multiple UPSERT queries run concurrently, they may all determine that the target document does not exist and then create it multiple times!!
> 
> Note that due to the non-atomicity of the lookup and insert/update/replace, even with a unique index there may be duplicate key errors or conflicts. But if they occur, the application/client code can execute the same query again
> 
> To prevent this from happening, one should a unique index to the lookup attribute(s). Note that in the cluster a unique index can only be created if it is equal to the shard key attribute of the collection or at least contains it as a part.
> 
> An alternative to making an UPSERT statement work atomically is to use the `exclusive` option to limit write concurrency for this collection to 1, which helps avoiding conflicts but is bad for throughput!

Shouldn't this be added as well?

mchacki:
Yes I agree, this should be included.

PR #538, opened 2020-08-31
add V8 usage metrics

jsteemann:
Docs PR for https://github.com/arangodb/arangodb/pull/12560 and https://github.com/arangodb/arangodb/pull/12561


dothebart:
hm, shouldn't these files be auto-generated? I don't know whether or not - but most probably in here they will be overwritten later on? Else, how will they make it into the swagger json?

Simran-B (approved):
It would be great if we could auto-generate it, but I'm not aware of such a facility.

PR #544, opened 2020-09-10
add AQL limits

jsteemann:
https://deploy-preview-544--zealous-morse-14392b.netlify.app/docs/3.7/aql/fundamentals-limitations.html

### DIFF
--- a/3.8/aql/fundamentals-limits.md
+++ b/3.8/aql/fundamentals-limits.md
@@ -18,4 +18,4 @@ The following limitations are known for AQL queries:
   function, or traversals working with collection sets (instead of named graphs).
 - Subqueries that are used inside expressions are pulled out of these
   expressions and executed beforehand. That means that subqueries do not
-  participate in lazy evaluation of operands.
+  participate in lazy evaluation of operands, like the ternary operator.
--- a/3.8/aql/fundamentals-limits.md
+++ b/3.8/aql/fundamentals-limits.md
@@ -13,9 +13,9 @@ The following limitations are known for AQL queries:
 - An AQL query cannot use more than _2048_ collections/shards.
 - It is not possible to use a collection in a read operation after
   it was used for a write operation in the same AQL query.
-- In the cluster, all vertex collections and collection that are accessed
-  dynamically via the `DOCUMENT` AQL function must be stated in the 
-  query's initial `WITH` statement.
+- In the cluster, all collections that are accessed dynamically must be stated in the 
+  query's initial `WITH` statement. These are either accesses via the `DOCUMENT` AQL
+  function, or traversals working with collection sets (instead of named graphs).
 - Subqueries that are used inside expressions are pulled out of these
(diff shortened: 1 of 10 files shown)


3.8/aql/fundamentals-limits.md
```diff
@@ -0,0 +1,21 @@
...
+description: Known limitations that apply in AQL queries
+---
+Known limitations for AQL queries
+=================================
+
+The following limitations are known for AQL queries:
+
+- An AQL query cannot use more than _1000_ result registers. One
+  result register is needed for every named query variable and for 
+  internal/anonymous query variables.
```
goedderz, line 12: As since 3.7 we can reuse a register after a variable is no longer needed, this sounds more pessimistic than is actually the case for most queries. I'm not sure though whether we want to go into more detail or not.

dothebart (approved):
btw, reading the `WITH` clause - we still have a feature request to be able to turn that on for single server ;)

3.8/aql/fundamentals-limits.md
```diff
@@ -0,0 +1,21 @@
...
+- An AQL query cannot use more than _2048_ collections/shards.
+- It is not possible to use a collection in a read operation after
+  it was used for a write operation in the same AQL query.
+- In the cluster, all vertex collections and collection that are accessed
+  dynamically via the `DOCUMENT` AQL function must be stated in the 
+  query's initial `WITH` statement.
```
goedderz, line 18: I think that _all vertex collections_ might be misleading. Perhaps something like
```suggestion
- In the cluster, all collections that are accessed dynamically must be stated in the 
  query's initial `WITH` statement. These are either accesses via the `DOCUMENT` AQL
  function, or traversals working with collection sets (instead of named graphs).
```
Maybe with links to `DOCUMENT` and https://www.arangodb.com/docs/stable/aql/graphs-traversals.html#working-with-collection-sets.

One other note: We have the `OPTION { vertexCollections }` since 3.7. These used could be used to add the specified collections to the transaction, but I don't know whether they are.

3.8/aql/fundamentals-limits.md
```diff
@@ -0,0 +1,21 @@
...
+- Subqueries that are used inside expressions are pulled out of these
+  expressions and executed beforehand. That means that subqueries do not
+  participate in lazy evaluation of operands.
```
goedderz, line 21:
```suggestion
  participate in lazy evaluation of operands, like the ternary operator.
```
As this is probably the most common pitfall, I'd add it, even though it is only one case.

3.8/aql/fundamentals-limits.md
jsteemann: Although you are right that some registers may be reused, I wouldn't document this as a "guaranteed" feature.
Queries using close to 1000 registers will suffer from various issues anyway.

Simran-B (approved):

PR #547, opened 2020-09-11
- added tash

neel:
added community C++ developed library tash (https://gitlab.com/neel.basu/tash)


Simran-B:
Thanks a lot for your contribution to ArangoDB @neel!

I wonder whether this counts as driver or integration. I would count a query build as integration, but it looks like you also implemented some functionality to communicate with the server?

Furthermore, for legal reasons we need you to fill out and sign our Contributor License Agreement (CLA). We use an Apache 2 CLA for ArangoDB, which can be found here: https://www.arangodb.com/documents/cla.pdf

neel:
Yes it communicates with the server through the HTTP API. However the project that I was working on then required complicated AQL. So simple creation, updation etc.. was not enough. So an AQL builder DSL was also written to generate the AQL and send that to the server and fetch the result set.
I'll send the CLA.

neel:
@Simran-B sent signed CLA as mail.

Simran-B:
Thanks again. As soon as the docs are published again, tash will beon the list.

PR #553, opened 2020-09-17
AQL operations: more headlines

Simran-B:
https://deploy-preview-553--zealous-morse-14392b.netlify.app/docs/3.8/aql/operations-update.html

**TODO:** Check statements in `aql/operations-collect.md`

> In order to make a result set unique, one can either use `COLLECT` or
`RETURN DISTINCT`. Behind the scenes, both variants create a *CollectNode*.
For both variants, the optimizer may try the <ins>sorted and the hashed variant of
`COLLECT`</ins>. The difference is therefore mainly <ins>syntactical</ins>, with
`RETURN DISTINCT` saving a bit of typing when compared to an equivalent `COLLECT`:

vs.

> `RETURN DISTINCT` does <ins>not change the order of results</ins>, whereas `COLLECT` sorts
them (regardless of the method, _sorted_ or _hash_) unless explicitly disabled
by the user with a subsequent `SORT null`

The last part of the sentence does not seem to be true, `COLLECT ... OPTIONS { method: "hash" } SORT null` does not return array elements in the original order.


Simran-B:
`RETURN DISTINCT` maintains order, whereas `COLLECT` does not (sorted or undefined order). Internally, there are four different implementations for CollectNode: hash, sorted, distinct and count. The latter two cannot be specified like `COLLECT ... OPTIONS { method: "..." }`.

PR #561, opened 2020-10-08
add testreport in progress

dothebart:


Simran-B:
I believe this should go to https://github.com/arangodb/documents/

PR #568, opened 2020-10-27
added aql analyzer description

Dronplane:
Added description for new analyzer type "aql"


Simran-B:
Merging for now, despite that some of the examples return unexpected things.

Internal references:

- arangodb/backlog#780
- arangodb/backlog#781
- arangodb/backlog#782

PR #571, opened 2020-10-30
feature/arangosearch-s2

gnusi:
https://github.com/arangodb/arangodb/blob/devel/tests/js/common/aql/aql-view-arangosearch-noncluster.js#L2200


Simran-B (approved):
Analyzer examples to be added later.

PR #572, opened 2020-11-03
mention restrictions for OneShard optimization

jsteemann:
Add extra "Limitations" section for OneShard feature


Simran-B (approved):

Simran-B:
DOCUMENT() function eligible in 3.7.5 and 3.8.0. Added remark for older versions.

PR #576, opened 2020-11-09
Automatically add "www-authenticate" headers to server HTTP 401 responses

jsteemann:
https://github.com/arangodb/arangodb/pull/13001


Simran-B (approved):
Also added to API changes, fixed capitalization of Www-Authenticate in some places and added a known issue for 3.5-3.7 (3.4 definitely affected as well, but EoL). No plans to backport according to Vadim.

PR #578, opened 2020-11-10
Foxx: graphql-sync is deprecated

Simran-B:
@pluma I think this doesn't go far enough yet. WDYT about moving the part about 0.12+ up, and move the legacy stuff to the bottom, under a separate headline?

### DIFF
--- a/3.5/foxx-reference-modules-graph-ql.md
+++ b/3.5/foxx-reference-modules-graph-ql.md
@@ -1,28 +1,51 @@
 ---
 layout: default
-description: GraphQL integration
+description: >-
+  The @arangodb/foxx/graphql module lets you create routers for serving GraphQL requests
+title: GraphQL integration for Foxx
 ---
 GraphQL integration
 ===================
 
 `const createGraphQLRouter = require('@arangodb/foxx/graphql');`
 
-Foxx bundles version 0.6 of the
-[`graphql-sync` module](https://github.com/arangodb/graphql-sync){:target="_blank"}, which is a
-synchronous wrapper for the official JavaScript GraphQL reference
-implementation, to allow writing GraphQL schemas directly inside Foxx.
-
-Additionally the `@arangodb/foxx/graphql` lets you create routers for serving
-GraphQL requests, which closely mimicks the behavior of the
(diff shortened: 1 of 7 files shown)


pluma:
Yes, I think it makes sense to invert the order here and only mention `graphql-sync` for backwards-compatibility. What we want new users to do is to bundle their own version of `graphql`.

3.5/foxx-reference-modules-graph-ql.md
```diff
@@ -1,29 +1,52 @@
...
+[`graphql-sync` module](https://github.com/arangodb/graphql-sync){:target="_blank"},
+which is a synchronous wrapper for the official JavaScript GraphQL reference
+implementation.
+
+{% hint 'warning' %}
+`graphql-sync` is a thin wrapper for old versions of `graphql`, allowing it
+to run in ArangoDB. This GraphQL server/schema implementation is deprecated
+and only shipped for backward compatibility. Version 0.12 and newer of the
+official `graphql` package can be used directly. New projects should bundle
+an own copy of this module: <https://www.npmjs.com/package/graphql>
```
pluma, line 47: sorry for not catching this earlier: "an own" reads weird. better: "their own"?

3.5/foxx-reference-modules-graph-ql.md
```diff
@@ -54,39 +77,9 @@ router.use('/graphql', createGraphQLRouter({
...
-});
-module.context.use(createGraphQLRouter({
-  schema: graphqlSchema,
-  graphiql: true,
-  graphql: graphql
-}))
-```
+For more information on `graphql-sync` see the
+[`graphql-js` API reference](http://graphql.org/docs/api-reference-graphql/){:target="_blank"}
+(note that `graphql-sync` never wraps results in promises).
```
pluma, line 82: also this aside about promises is irrelevant as graphql no longer wraps results in promises unless the resolution method is defined as async, so this just works now. it doesn't do any harm but it's also not informative.

Simran-B:
Addressed in this commit: https://github.com/arangodb/docs/commit/d2c7db83b760c170c7e6aa3e08144b5215108b24

PR #584, opened 2020-11-17
Respect --network.io-threads startup option

jsteemann:
Make the number of network I/O threads properly configurable via the startup option --network.io-threads. This option existed before, but its configured value was effectively clamped to a value of 1.

https://github.com/arangodb/arangodb/pull/13051

Will be fixed in v3.6.10, v3.7.5 and v3.8.devel


Simran-B (approved):
Added known issue retrospectively

PR #585, opened 2020-11-18
Canonical redirects

Simran-B:
Before:
- `/3.7/ -> /3.7/` ✔️ 
- `/stable/ -> /3.7/` ❌ 

After:
- `/3.7/ -> /stable/` ❌ 
- `/stable/ -> /stable/` ✔️ 

Desired:
- `/3.7/ -> /3.7/` ✔️ 
- `/stable/ -> /stable/` ✔️ 

Problem: `/stable` is symlinked and can only forward to either `/3.7` or `/stable`. We would need to generate the files for stable and devel version separately (or at least the redirect files).


Simran-B:
Too much effort for too little gain. Closing.

PR #587, opened 2020-11-23
ArangoSearch / Analyzers restructure

Simran-B:
Goals:
- Separate ArangoSearch and Analyzers
- Separate learning content from reference content (Views page should be reference only)
- Make the main ArangoSearch page a practical introduction
  (explaining things like Analyzer context here instead of in the reference docs of the ArangoSearch AQL functions, i.e. make it more self-contained)
- Replace examples on Examples page with better and more examples by turning the examples page into a collection of use cases (the PR only starts this, to be extended over time)


3.8/aql/functions-arangosearch.md
```diff
@@ -899,7 +849,8 @@ Scoring functions return a ranking value for the documents found by a
 the search expression the higher the returned number.
 
 The first argument to any scoring function is always the document emitted by
-a `FOR` operation over an ArangoSearch View.
+a `FOR` operation over an ArangoSearch View. Scoring functions can be used in
+`SORT` and `RETURN` operations only.
```
gnusi, line 853: that's not true, score value can be used everywhere

gnusi (changes requested):

Simran-B:
Tasks:
- Split _Search functions_
  - **Context functions**: return expression result and affect SEARCH context (ANALYZER, BOOST)
  - **Filter functions**: meanwhile usable outside of SEARCH expressions (except PHRASE, EXISTS), can be accelerated with View index(es)
- Search context needs to be properly explained
- Using ArangoSearch functions with or without a View/the SEARCH keyword needs to be explained (index utilization etc.). Needs refactoring of existing docs because it is assumed that the first argument is an attribute path, not a literal.
- Revert change from "returns nothing" to "returns bool" where the return value is inaccessible for the user?
- Separate ArangoSearch (IResearch, full-text search and scoring, View type) from Analyzers (text processing)
- Sync HTTP and Manual content
- Improve ArangoSearch Views page (drop concepts, give short example at the beginning, potentially move introduction to parent page and keep only reference documentation separate)
- Remove cross-references to string functions? (TOKENS, NGRAM_SIMILARITY, NGRAM_POSITIONAL_SIMILARITY)

Simran-B:
Meeting notes 2021-03-30:
- List all operators somewhere. Does not have to be on the main page, but it should at least reference another page with a full list.
- Rework main page to be a tutorial by moving the performance optimization section to its own page (it's more of an advanced guide)
- Split examples up into meaningful pages:
  - (separate page for each?!)
    - Exact value
    - Prefix
    - Auto-complete / edge-ngrams / ???
    - Case-insensitive
    - Wildcard
  - **Ranking**
    - scoring functions
    - boost context
    - custom score formula
    - norm Analyzer feature:
      TFIDF/BM25, performance implications, longer text with the same number of mentions of term = equally relevant (and norm tries to distribute the score accordingly)
    - known issue: _Score values evaluated by corresponding score functions (BM25/TFIDF) may differ in single-server and cluster with a collection having more than 1 shard_)
  - **Performance optimization**
    - `primarySort`
    - `storedValues`
    -  norm Analyzer feature performance implications?
  - **Fuzzy search**
  - **Geo search**
  - **Testing** ?
    We would mention `waitForSync` here. Needs to warn about permanent inconsistencies in Views index if transactions are rolled back.
- Check with Heiko from where the IMDB dataset was taken and how it was processed
- Create a release in example-datasets repo to make it easier to download the dataset

Simran-B:
`inBackground` is an option, not a property (does not get persisted, needs to be set when adding a link in order to make indexing run in background) - https://github.com/arangodb/arangodb/issues/13414

PR #589, opened 2020-11-23
Fix "Contribute" link for "Oasis" pages.

likemusic:
`oasis` directory is symlinked to version directory (for example `3.7`). Therefore `page.path` starts with version directory, like `3.7/oasis/organizations.md` instead of real path `oasis/organizations.md` that should be used to generate link for "Contribute"-button.

I'm not sure that my fix implementation is the right way for this case. Seems, that it would be better to have custom filter to resolve symlink, but I don't not if it is possible with Jekyll.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -4,7 +4,7 @@
   {%- include head.html %}
 </head>
 <body>
-  {% include topnav.html %} 
+  {% include topnav.html %}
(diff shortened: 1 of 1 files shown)


Simran-B:
Implemented a generic fix in #592, which adds a filter to resolve symlinks.

PR #597, opened 2020-12-10
Known issue: db not found when slow queries are accessed after db creation

Simran-B:
For 3.7, the fix may get delayed to 3.7.7!


KVS85 (approved):
LGTM.

Simran-B:
Will be fixed in v3.7.6

PR #599, opened 2020-12-17
Clarifications around the WITH operation

Simran-B:
Distinguish between MMFiles and RocksDB, only the latter in 3.7+ docs. Explain how deadlocks can happen.


Simran-B:
Mixing OneShard and sharded collections in a query is fine, the optimizer does not push the DOCUMENT() function call down to a DB-Server at all for databases with sharding `""` / `"flexible"`, see #756

PR #605, opened 2021-01-02
documentation for DATE_TIMEZONE(S) and zoneinfo switch

coronabytes:
Documentation for the new AQL functions in PR [#13334](https://github.com/arangodb/arangodb/pull/13334) / [#13464](https://github.com/arangodb/arangodb/pull/13464)
and  [#13320](https://github.com/arangodb/arangodb/pull/13320) 

DATE_TIMEZONE
DATE_TIMEZONES
DATE_UTCTOLOCAL
DATE_LOCALTOUTC


coronabytes:
I see were closing in on this.

I'm having some problems with the formatting on the return object from DATE_UTCTOLOCAL and DATE_LOCALTOUTC.
Is there any other sample in the docs where you describe a 'complex' return object.

Simran-B:
Sorry that it took so long.
There aren't that many functions which return objects. Most of them are so simple that they are described in a single sentence. There's one notable exception however:
https://www.arangodb.com/docs/stable/aql/functions-miscellaneous.html#schema_validate
The formatting/notation is admittedly a bit strange, but I can't think of a good alternative. I'm open to suggestions.

Simran-B:
Let me know if you can think of a better way to format complex return values.

PR #609, opened 2021-01-20
Pregel: Collection sharding in cluster

Simran-B:
Needs sharding guarantees of SmartGraphs so that the outgoing edges are placed on the same DB-Server as the vertex.

Preview: https://deploy-preview-609--zealous-morse-14392b.netlify.app/docs/3.8/graphs-pregel.html#prerequisites


vgasiunas:
Looks good. Thanks.

PR #615, opened 2021-02-01
Add section about missing locale on Linux / language mismatch on startup

Simran-B:
Server fails to start on some systems:

```
FATAL [7ef60] {config} specified language 'en_US' does not match previously used language ''
```

Related issues:
- https://github.com/arangodb/arangodb/issues/12501
- https://github.com/arangodb/arangodb/issues/8332

It's unclear how it gets into this situation and whether starting the server initially with an explicit `default-language` would help in such cases. We should also add information about how to switch it (dump & restore) and potentially improve the server error message in cases where there is no system locale.

### DIFF
--- a/3.6/installation-linux-osconfiguration.md
+++ b/3.6/installation-linux-osconfiguration.md
@@ -13,6 +13,36 @@ easily applied by making use of a script. Please refer to the page
 ready-to-use examples.
 {% endhint %}
 
+System Locales
+--------------
+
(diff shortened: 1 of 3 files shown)


Simran-B:
Netlify's `[skip ci]` is just broken...

Simran-B:
While there seem to be more issues with the locale, e.g. with the Debian package and under Windows, this remark should at least help some Linux users.

PR #623, opened 2021-02-22
adjust docs for changed splice-subqueries optimizer rule behavior

jsteemann:
Adjust the docs for 3.8 so they mention the optimizer rule `splice-subqueries` cannot be turned off anymore.

### DIFF
--- a/3.8/release-notes-upgrading-changes38.md
+++ b/3.8/release-notes-upgrading-changes38.md
@@ -301,6 +301,18 @@ FOR status IN `Window`
   RETURN status.open
 ```
 
+## Subqueries
+
+The AQL optimizer rule `splice-subqueries` was introduced in ArangoDB 3.6 to
+optimize most subqueries, and it was extended in 3.7 to work with all types
+of subqueries. It was always turned on by default, but it still could be 
+deactivated manually using a startup option (`--query.optimizer-rules`).
+
+In ArangoDB 3.8, the optimizer rule `splice-subqueries` is now required for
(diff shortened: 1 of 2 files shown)


3.8/release-notes-upgrading-changes38.md
```diff
@@ -301,6 +301,18 @@ FOR status IN `Window`
   RETURN status.open
 ```
 
+## Subqueries
+
+The AQL optimizer rule `splice-subqueries` was introduced in ArangoDB 3.6 to
+optimize most subqueries, and it was extended in 3.7 to work with all types
+of subqueries. It was always turned on by default, but it still could be 
+deactivated manually using a startup option (`--query.optimizer-rules`).
```
Simran-B, line 309:
```suggestion
deactivated manually using a startup option (`--query.optimizer-rules`) or
disabled for individual queries via the `optimizer.rules` query option.
```

3.8/release-notes-upgrading-changes38.md
```diff
@@ -301,6 +301,18 @@ FOR status IN `Window`
...
+
+In ArangoDB 3.8, the optimizer rule `splice-subqueries` is now required for
+subquery execution, and cannot be turned off. Trying to disable it via the 
+mentioned startup option has no effect, as the optimizer rule will always 
+run for queries containing subqueries.
```
Simran-B, line 314:
```suggestion
mentioned startup option or query option has no effect, as the optimizer rule
will always run for queries containing subqueries.
```

Simran-B (approved):

PR #624, opened 2021-02-22
adjust memory limit formula for small installations

jsteemann:
@simran-b: as discussed, here is the adjustment for the memory usage limits.

This is the docs PR for https://github.com/arangodb/arangodb/pull/13595

### DIFF
--- a/3.8/release-notes-upgrading-changes38.md
+++ b/3.8/release-notes-upgrading-changes38.md
@@ -191,8 +191,9 @@ option `--query.memory-limit` from previously `0` (meaning: no limit) to a
 dynamically calculated value. The per-query memory limits defaults are now:
 
 ```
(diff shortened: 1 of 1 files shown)


3.8/release-notes-upgrading-changes38.md
```diff
@@ -191,8 +191,9 @@ option `--query.memory-limit` from previously `0` (meaning: no limit) to a
 dynamically calculated value. The per-query memory limits defaults are now:
```
Simran-B, line 191:
```suggestion
dynamically calculated value. The per-query memory limit defaults are now:
```

Simran-B (approved):

Simran-B:
@jsteemann Thanks!

PR #634, opened 2021-03-13
Fix slower-than-necessary arangoimport behavior

jsteemann:
https://github.com/arangodb/arangodb/pull/13704

Implementation detail: the rate limiter previously started out with 1 MB/s and then dynamically increased the throughput by 20% every 10 seconds. Now it starts with 8 MB/s (new default `--batch-size`) and dynamically increases by 25% every 2 seconds.


Simran-B (changes requested):
This needs to be changed accordingly: https://github.com/arangodb/docs/blob/bug-fix/fix-arangoimport-slowness/3.8/programs-arangoimport-details.md#automatic-pacing-with-busy-or-low-throughput-disk-subsystems

Simran-B (approved):

PR #651, opened 2021-03-24
Added an option to lock down /_admin/cluster

jsteemann:
Docs PR for https://github.com/arangodb/arangodb/pull/13816


Simran-B:
Partially also documented via #652

PR #654, opened 2021-03-29
Feature/stopwords analyzer

Dronplane:
Add description and release notes for the stopwords analyzer


gnusi (approved):

Simran-B:
- Hex-encode UTF-8 string with JS in browser:
  `Array.from(new TextEncoder().encode("\tö"), byte => byte.toString(16).padStart(2, "0")).join("")`
- Node.js / arangosh: `Buffer("\tö").toString("hex")`
- AQL: `RETURN TO_HEX("\tö")`

PR #661, opened 2021-04-01
add more deprecated things

jsteemann:


Simran-B (approved):
Partially applied additions to 3.7. Removed the clusterInventory entry - this page should be limited to higher-level stuff, for everything else there are the incompatible changes page and the CHANGELOG.

PR #662, opened 2021-04-01
mention support for gzip-compressed input files

jsteemann:


Simran-B (approved):
I can confirm that arangoimport accepts .gz files in 3.7.1 and 3.6.0, possibly even older versions.

PR #664, opened 2021-04-05
add returnType description for AQL analyzer

Dronplane:
adds descriptions for new Aql analyzer parameter

https://github.com/arangodb/arangodb/pull/13894

### DIFF
--- a/3.8/arangosearch-analyzers.md
+++ b/3.8/arangosearch-analyzers.md
@@ -489,8 +489,16 @@ be run on DB-Servers in case of a cluster deployment. User-defined functions
 are not permitted.
 
 The input data is provided to the query via a bind parameter `@param`.
-It is always a string. The query result should be a string, `null` or
-an array of strings (and `null`s).
+It is always a string. The query result will be converted according to
+`returnType` parameter value. 
+
+{% hint 'information' %}
+If `returntType` is `"number"` or `"bool"` there is no need 
+to use custom analyzer via `ANALYZER()` function in your queries. Numerics and booleans
+will be available the same way as if these values were directly in indexed documents.
+{% endhint %}
+
+
 
 The *properties* allowed for this Analyzer are an object with the following
 attributes:
@@ -512,6 +520,16 @@ attributes:
   performance.
 - `memoryLimit` (integer): memory limit for query execution in bytes.
   (default is 1048576 = 1Mb) Maximum is 33554432U (32Mb)
+
+- `returnType` (string): Type of returned tokens.
+  - `"string"` (default) to mark emitted tokens as strings
+  - `"number"` to mark emitted tokens as numerics
+  - `"bool"` to mark emitted tokesn as bools.
+
+  If type does not match
+  actual value type returned by `queryString` implicit type conversion is
+  done according to AQL type conversion rules see [TO_BOOL()](./aql/functions-type-cast.html#to_bool),
+  [TO_NUMBER()](./aql/functions-type-cast.html#to_number), [TO_STRING()](./aql/functions-type-cast.html#to_string)
 
 **Examples**
 
--- a/3.8/arangosearch-analyzers.md
+++ b/3.8/arangosearch-analyzers.md
@@ -495,7 +495,7 @@ It is always a string. The query result will be converted according to
 {% hint 'information' %}
 If `returntType` is `"number"` or `"bool"` there is no need 
 to use custom analyzer via `ANALYZER()` function in your queries. Numerics and booleans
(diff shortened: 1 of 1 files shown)


3.8/arangosearch-analyzers.md
```diff
@@ -489,8 +489,16 @@ be run on DB-Servers in case of a cluster deployment. User-defined functions
...
 The input data is provided to the query via a bind parameter `@param`.
-It is always a string. The query result should be a string, `null` or
-an array of strings (and `null`s).
+It is always a string. The query result will be converted according to
+`returnType` parameter value. 
+
+{% hint 'information' %}
+If `returntType` is `"number"` or `"bool"` there is no need 
+to use custom analyzer via `ANALYZER()` function in your queries. Numerics and booleans
+will be available the same way as if these values were directly in the indexed documents.
```
Simran-B, line 498: Does this try to explain the following comment from the feature PR?

> automatically apply internal numeric/boolean analyzers for tokenizing corresponding values

It's the first time I hear about internal number/bool analyzers. Do we really need to explain this if it's hidden from the user and just works as expected?

Or does this refer to something completely different? That users should not create AQL Analyzers only to index plain numbers or booleans, because that works out of the box with any Analyzer?

3.8/arangosearch-analyzers.md
```diff
@@ -489,8 +489,16 @@ be run on DB-Servers in case of a cluster deployment. User-defined functions
 are not permitted.
 
 The input data is provided to the query via a bind parameter `@param`.
-It is always a string. The query result should be a string, `null` or
-an array of strings (and `null`s).
+It is always a string. The query result will be converted according to
+`returnType` parameter value.
```
Simran-B, line 493: Does the implementation literally convert the AQL result to _returnType_?
For example, if I `RETURN true` but set `returnType: "string"`, will it end up as string `"true"`?
And if I `RETURN "true"`  and set `returnType: "bool"`, will it end up as boolean `true`?

3.8/arangosearch-analyzers.md
```diff
@@ -513,6 +521,16 @@ attributes:
 - `memoryLimit` (integer): memory limit for query execution in bytes.
   (default is 1048576 = 1Mb) Maximum is 33554432U (32Mb)
 
+- `returnType` (string): Type of returned tokens.
```
Simran-B, line 524: Tokens (plural) - does that mean the AQL query can return an array where each element is of the indicated type?

3.8/arangosearch-analyzers.md
Dronplane: Exactly

3.8/arangosearch-analyzers.md
Dronplane: No, it means, analyzer could emit arbitrary number of tokens.

3.8/arangosearch-analyzers.md
Dronplane: Well, by internal conversion rules I wanted to tell about corresponding type conversion functions e.g. TO_STRING

Simran-B: Okay, so if I use an AQL Analyzer with the "number" returnType for instance, then I don't have to set a context Analyzer in the query to match that AQL Analyzer. I can simply compare the value directly, like `SEARCH doc.value == 42` (defaults to identity, but Analyzer doesn't matter because processing is skipped for booleans and numbers). Using `SEARCH ANALYZER(doc.value == 42, "my_aql_analyzer_with_number_return_type")` would work as well, but is unnecessary - is that correct?

Dronplane: Yes

3.8/arangosearch-analyzers.md
Simran-B: So `RETURN [1, 2]` will not work (result: `[ [ 1, 2 ] ]`) but `FOR val IN [1, 2] RETURN val` will? (result: `[1, 2]`)?
In other words, each top-level element of the result is a token, or am I getting this totally wrong?

Dronplane: Actually RETURN [1,2] will work. All will depend on  `returnType` value. E.g. if it is string than token will be "[1,2]". And yes you are right each value emitted by RETURN node is considered as transformation function input. If it is array or object - it will be converted as one entity, not iterated.

Simran-B:
Examples to be added later when https://arangodb.atlassian.net/browse/BTS-357 is fixed

PR #675, opened 2021-04-12
renamed minCells to maxCells [skip netlify]

rashtao:


Simran-B (approved):
While there is also a minCells property, it doesn't appear to be exposed after a quick search in the IResearch code.

PR #680, opened 2021-04-14
docs PR for https://github.com/arangodb/arangodb/pull/13958

jsteemann:


Simran-B (commented):
Unclear whether WITH is also required for `DOCUMENT()` in some cases in cluster

jsteemann:
@Simran-B : note that I just added a backport for 3.7.12.

There is no need to use `WITH` with the `DOCUMENT` AQL function with RocksDB.

PR #683, opened 2021-04-15
Add --ldap2.* page for arangod

Simran-B:
Added an exception for `ldap2` to link to `ldap` subpage. It's a straight-up copy to specify an alternative LDAP server.

Added section about `--ldap*.responsible-for`.
There's a typo in the description BTW, missing an s in usernames: `A file of username ...` (in enterprise code)

Please review the content, especially whether ldap2 acts as a fail-over in case the primary LDAP server isn't reachable but also if the primary isn't responsible for a user (it's just an assumption).

- Overview
  https://deploy-preview-683--zealous-morse-14392b.netlify.app/docs/3.6/programs-arangod-options.html#ldap2-secondary-server
- New section
  https://deploy-preview-683--zealous-morse-14392b.netlify.app/docs/3.6/programs-arangod-ldap.html#secondary-server-options-ldap2

### DIFF
--- a/3.6/programs-arangod-ldap.md
+++ b/3.6/programs-arangod-ldap.md
@@ -57,7 +57,6 @@ actual list of roles of a user is derived from the LDAP server.
 See [Roles search](#roles-search) below for details about method (b)
 and for the associated configuration options.
 
-
 Fundamental options
 -------------------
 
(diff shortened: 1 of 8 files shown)


fceller (approved):
LGTM

mchacki:
After we talked about it, I now understand what the Documentation tries to say, but without our talk the Documentation would not have helped me.

All this is unnecessarily confusing:
1) The naming suggest LDAP version 2 not some secondary LDAP server
2) My first understanding of it was that the second one is just an identical failover server. Which is also not true.
3) The real thing is that you have two servers running that each have some potentially different configuration and users, and we basically check both servers, unless there is restrictions given by some startup parameter, if the user is in one of them she can log in.

So for documentation I think we need the Use-case (Which is perfectly valid) to be described. And some examples on how to actually use it.

Simran-B:
1. That's unfortunate indeed but cannot be changed anymore as far as I understand.
2. I was told otherwise. You can use the ldap2 options with identical settings except for a different server to use it as a failover server.
3. The second way you can utilize ldap2 is to make the primary server responsible for only some users and let the secondary handle the rest (or, if neither is responsible, the user will not be authenticated).

I agree that a worked example would be useful, but I have no idea how to set up two LDAP servers to test it. Can you help with that?

PR #688, opened 2021-04-19
improve parallelism capabilities of arangorestore

jsteemann:
https://github.com/arangodb/arangodb/pull/14010

### DIFF
--- a/3.8/programs-arangorestore-examples.md
+++ b/3.8/programs-arangorestore-examples.md
@@ -238,6 +238,18 @@ The following factors affect speed of _arangorestore_ in a Cluster:
   on multiple _Coordinators_ at the same time. Depending on your specific
   case, parallelizing on multiple _Coordinators_ can still be useful even
   when the `--threads` option is in use (from v.3.4.0).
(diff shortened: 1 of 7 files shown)


Simran-B (changes requested):
Please incorporate `--envelope false` into the fast cluster restore article and script.

jsteemann:
> Please incorporate `--envelope false` into the fast cluster restore article and script.
> 
> [`feature`/improve-restore-performance/3.8/programs-arangorestore-fast-cluster-restore.md](https://github.com/arangodb/docs/blob/feature/improve-restore-performance/3.8/programs-arangorestore-fast-cluster-restore.md)

IMHO scripts like "fast cluster restore" shouldn't even exist.
It should be fast out of the box, without the requirement to run custom scripts and take extra steps.
Apart from that, arangorestore does not have the `--envelope` option, but only arangodump has. arangorestore can pick up whatever dump exists and use it.

Simran-B:
I don't disagree but it still seems to have relevance as it talks to multiple Coordinators instead of one and there's something about a temporary replication factor of 1. I added a remark about the non-enveloped format.

PR #694, opened 2021-05-11
Remove Simple Queries, JS Traversal docs and non-Rest DocuBlocks

Simran-B:
- Removed docs about **Simple Queries** and **JS Traversal**.
  Both are deprecated since 3.4.0, should no longer mention them. Code will be removed in a future version.
- Removed docs for the JS API methods `cursor.setBatchSize()` and `cursor.getBatchSize()` from docs.
 The batch size methods were only supported for Simple Queries and the `query.execute()` method also accepts a batch size. For AQL queries, we only support an upfront cursor option.
- Removed docs for the JS API methods `collection.range()` and `collection.closedRange()`.
  Range queries are better done with AQL.
 - Inlined non-Rest DocuBlocks since they were only included once (with one exception), also deleted some obsolete content (associated with DocuBlocks).
- Removed some 2.8 -> 3.0 remarks. The changes are partially documented in the release notes.
  What does not appear to be documented is the addition of the parameters `silent`, `returnOld` and `returnNew`, but since the parameters themselves are documented this should be okay.

Depends on:
- https://github.com/arangodb/arangodb/pull/14179
- https://github.com/arangodb/arangodb/pull/14171
- https://github.com/arangodb/arangodb/pull/14245


jsteemann (approved):
LGTM.
Thanks a lot!

Simran-B:
https://jenkins01.arangodb.biz/view/Documentation/job/arangodb-ANY-examples/755/

joerg84 (approved):

PR #699, opened 2021-05-23
Fix arithmetic example

cugu:


Simran-B:
Thanks, I also fixed another mistake in the text and examples (1 / 0 = null, not 0).

PR #701, opened 2021-05-23
Fix object / document comparison

cugu:


Simran-B:
Oh wow, well spotted. It even says a few lines above:

> If one of the objects / documents does not have an attribute with the sought name, its attribute value is considered to be *null*.

I can confirm that `{ } == { "a": null } // true` in version 2.5 and 3.9.devel.

cugu:
This happens when you create unit tests out of the documentation ;-) #699 and #702 showed there as well.

Simran-B:
Thank you for reporting and fixing the docs!

Issue #417, opened 2020-04-18
Add log topic libiresearch to HTTP docs


Simran-B:
_David Costa Faidella in community slack:_

I can confirm level = libiresearch=fatal  changes the level so errors stop appearing (another story if it still generating them)

then it causes a warning on consolidation

`2020-04-18T17:06:49Z [3064] WARNING [bce4f] {arangosearch} error while consolidating arangosearch link '157': 4 failure while executing consolidation policy '{\n  "type" : "tier",\n  "segmentsBytesFloor" : 2097152,\n  "segmentsBytesMax" : 5368709120,\n  "segmentsMax" : 10,\n  "segmentsMin" : 1,\n  "minScore" : 0\n}' on arangosearch link '157' run id '331711769392'`

which is ignored by increasing arangosearch log level to error

Issue #296, opened 2020-01-13
ArangoSearch: FILTER/SEARCH OPTIONS { conditionOptimization: … } & 3.7 features


Simran-B:
The `conditionOptimization` option controls how conditions in SEARCH/FILTER operations are processed query.

https://github.com/arangodb/arangodb/pull/10781

If the option is omitted or set to `'auto'` then all conditions are converted to [disjunctive normal form](https://en.wikipedia.org/wiki/Disjunctive_normal_form) (DNF). Numerous optimizations can then be applying, like removing "redundant" / overlapping conditions (say if we have condition `a > 10 AND a > 5`, then we can leave out `a > 10` as it is included by `a > 5`).

However, converting to DNF takes time itself and DNF may become quite complex (e.g. customer tried more than 7000 conjunctions, but even 5-10 nested conditions should be enough to produce dozens of conjunctions; OR is also affected). In these corner cases it is faster to just search the index as is. This option controls this behavior.

Public modes:
- `auto`
- `none` (do not convert to DNF)

Internal only:
- `nodnf`
- `noneg`

It is currently enabled for ArangoSearch Views only (SEARCH operation). DNF is used for determining indexes for collections, which means it needs more work to implement it for FILTER as well.

Simran-B:
Example query with 7k conjunctions in ticket https://arangodb.atlassian.net/browse/DEVSUP-414

Also see https://github.com/arangodb/backlog/issues/596

Simran-B:
Another new feature: `LEVENSHTEIN_MATCH()`
https://github.com/arangodb/arangodb/pull/10895#pullrequestreview-344286458

Simran-B:
Note: conditionOptimization was backported to 3.6.2 https://github.com/arangodb/arangodb/pull/11038

Simran-B:
PHRASE() can skip empty arrays: https://github.com/arangodb/arangodb/pull/10954

Simran-B:
Added some docs about conditionOptimization to https://github.com/arangodb/docs/pull/402

Issue #222, opened 2019-11-14
ArangoSearch refactor (part 3)


Simran-B:
_Moved from https://github.com/arangodb/docs/pull/150 to here:_

### Open questions

- What does the `norm` Analyzer feature do?
  > **norm**: the field normalization factor

  → evens out scoring to make things fairer, so that ["bar","bar","bar"] vs. "bar" get more similar or even identical values. TFIDF has no norm. by itself, but our implementation adds it. For BM25 it is not meaningful to disable norm. Leaving the feature off for a custom Analyzer disables normalization – only if user desires so! Recommended to use it.

- Are the Analyzer features only necessary for the PHRASE() function?
  > For example the text type will produce frequency + norm + position and the PHRASE() AQL function requires frequency + position to be available.

  Does "produce" here mean that if the features are enabled then the Analyzer will compute this metadata, otherwise it won't?

  → PHRASE() is the only function so far. `norm` is not necessary and only controls scoring behavior (see above)

- `primarySort` does not round-trip `direction: "asc"/"desc"` but converts it to `asc: true/false`, will it stay this way? Should we recommend to use `asc` instead?

→ both syntaxes supported so far, we might want to stop advertising `direction` and recommend `asc` instead, then remove the former in 4.0.0

- `FOR d IN myView SEARCH ... OPTIONS {waitForSync:true} RETURN d` - is this "hidden" option supposed to be documented?

  → Do not advertise, only for unit tests. Flushes internal buffers, creates new segments etc. However, it could be used in example generation, as long as OPTIONS is hidden. May do the following:
  - insert docs
  - `FOR doc IN someView OPTIONS { waitForSync: true } LIMIT 1 RETURN true` (hidden, should block until View has indexed docs)
  - AQL query

- From 3.4 HTTP docs:
  > view can't be created with the links. Please use PUT/PATCH for links management.
  Was that the reason for the separation of View Definition/Modification and View / Link Properties?

  → They can be created with links, but adding links later can be better as it's not an atomic operation and some errors may occur. We might want to suggest users to always verify whether the links work (correct collection names, Analyzers, everything accessible etc.)

- was the `lookahead` property of Views removed?

  → tier policy, removed in 3.5

- Does any encoding other than UTF-8 in the locale make any sense in ArangoDB? (What encodings do we accept in general, and are they always normalized?)

[...]

PR #705, opened 2021-05-26
ArangoSearch clarifications

Simran-B:
- BTS-399: An n-gram Analyzer with min == max is expected by NGRAM_MATCH(), it isn't just a recommendation (score will be lower than expected otherwise)
- Remove `countApproximate` option from 3.7 docs (added only in 3.8)
- Approximate count (`countApproximate: "cost"`) is affected by the eventually consistent nature of Views, i.e. it is not precise if documents were recently changed because everything View-related is lagging behind
- Remove erroneous link JACCARD() function . Proximity search is PHRASE() with skip tokens (wildcards)
- Fix headline level of STARTS_WITH()
- The SEARCH `conditionOptimization` option was introduced in 3.6.2, not 3.7.0
- Replace absolute with relative query execution time in `EXISTS()` example
- PHRASE() skips empty arrays since 3.6.1

### DIFF
--- a/3.6/aql/functions-arangosearch.md
+++ b/3.6/aql/functions-arangosearch.md
@@ -437,6 +437,20 @@ It is the same as the following:
 FOR doc IN myView SEARCH PHRASE(doc.title, "quick", 1, "fox", 0, "jumps", "text_en") RETURN doc
 ```
 
+Empty arrays are skipped:
+
+```js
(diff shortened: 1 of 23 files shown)


Simran-B:
There are still some open items from the following issues, but will address them separately:
- https://github.com/arangodb/docs/issues/296
- https://github.com/arangodb/docs/issues/222
- https://github.com/arangodb/docs/issues/417

Dronplane (approved):
LGTM

PR #711, opened 2021-06-02
fix stopwords description

Dronplane:
add hex parameter for stopwords analyzer


Simran-B:
Will leave this open until 3.8.1 when we plan to re-add the stopwords Analyzer to the docs.

Simran-B (approved):
Actually, it's easier to comment out the `stopwords` Analyzer-related content instead of removing the content and later reverting that removal.

PR #717, opened 2021-06-11
Add fuzzy search examples

Simran-B:
Based on Lunch Break video

https://deploy-preview-717--zealous-morse-14392b.netlify.app/docs/3.7/arangosearch-fuzzy-search.html


gnusi:
@Simran-B LGTM, but would be nice to mention that similarity also affects overall score `BM25`/`TFIDF`

PR #731, opened 2021-07-15
fix login example

WhileOneK:
In the example above the authentication result is stored in the password field.


Simran-B:
Thanks a lot for your contribution to ArangoDB!

WhileOneK:
Thank you. I have send the cla pdf to the email like requested.

joerg84:
CLA received, thanks for the contribution!

@Simran-B please proceed

Simran-B (approved):
The example was apparently inspired by the dependencies guide, which names the attribute `authData` that stores the hashed password.

Thanks for fixing this!

PR #742, opened 2021-08-03
added general docu about hybrid smart graphs

hkernbach:
WIP

### DIFF
--- a/3.9/graphs-smart-graphs.md
+++ b/3.9/graphs-smart-graphs.md
@@ -71,6 +71,17 @@ overhead. The more subgraphs are touched the more network cost will apply.
 However the overall performance is never worse than the same query using a
 General Graph.
 
+Benefits of Hybrid SmartGraphs
+-------------------------------
+
+Hybrid SmartGraphs are capable of using Satellite Collections within their graph
+definition. Therefore, defined EdgeDefinitions between SmartCollections and
+SatelliteCollections can be created. As Satellite Collections are globally
+replicated to each participating DB-Server, graph traversals, shortest path,
+weighted traversals and k-shortest-paths queries can be executed partially
+locally on each DB-Server. This means that query execution is guaranteed to be
+fully local whenever data from the Satellite part is required.
+
 Benefits of Disjoint SmartGraphs
 -------------------------------
 
@@ -82,6 +93,13 @@ prohibits edges between vertices with different `smartGraphAttribute` values.
 This ensures that graph traversals, shortest path, and k-shortest-paths queries
 can be executed locally on a DB-Server, achieving improved performance for
 these type of queries.
+
+Benefits of Hybrid Disjoint SmartGraphs
+-------------------------------
+
+Hybrid Disjoint SmartGraphs benefit of all advantages which are present in
+Hybrid SmartGraphs. The rules of Disjoint SmartGraphs, described in the chapter
+above, stay true.
 
 Getting started
 ---------------


3.9/graphs-general-graphs-management.md
```diff
@@ -374,11 +376,13 @@ graph with different *from* and/or *to* collections an error is thrown.
...
 Modify a relation definition:
 
-`graph_module._editEdgeDefinitions(edgeDefinition)`
+`graph_module._editEdgeDefinitions(edgeDefinition, options)`
 
 - `edgeDefinition` (object):
   The edge definition to replace the existing edge definition with the same
   attribute *collection*.
+- `options` (object):
+  Additional options related to the edgeDefinition itself.
```
mchacki, line 385: Can we add a link to the Options defined later, stating what values are allowed here?
If not I think we need to copy the "satellite" sentence to here as well, otherwise it is not clear what to do from a user-perspective. Searching for "options" is kind of hopeless.

3.9/graphs-general-graphs-management.md
```diff
@@ -343,10 +343,12 @@ You can still add, delete or modify edge definitions and vertex collections.
 
 Add another edge definition to the graph:
 
-`graph._extendEdgeDefinitions(edgeDefinition)`
+`graph._extendEdgeDefinitions(edgeDefinition, options)`
 
 - `edgeDefinition` (object):
   The relation definition to extend the graph
+- `options` (object):
+  Additional options related to the edgeDefinition itself.
```
mchacki, line 351: Can we add a link to the Options defined later, stating what values are allowed here?
If not I think we need to copy the "satellite" sentence to here as well, otherwise it is not clear what to do from a user-perspective. Searching for "options" is kind of hopeless.

3.9/graphs-smart-graphs.md
```diff
@@ -71,6 +71,17 @@ overhead. The more subgraphs are touched the more network cost will apply.
...
 
+Benefits of Hybrid SmartGraphs
+-------------------------------
+
+Hybrid SmartGraphs are capable of using Satellite Collections within their graph
+definition. Therefore, defined EdgeDefinitions between SmartCollections and
+SatelliteCollections can be created. As Satellite Collections are globally
+replicated to each participating DB-Server, graph traversals, shortest path,
+weighted traversals and k-shortest-paths queries can be executed partially
+locally on each DB-Server. This means that query execution is guaranteed to be
```
mchacki, line 82:
```suggestion
locally on each DB-Server. This means a larger part of the query can be executed
```

We should not "guarantee" this here. There may be some code pieces that will not stay local.
e.g. we may introduce an upper limit of Operations per SmartGraph Request to improve LIMIT calls.

mchacki (changes requested):
Some suggestions.

I am missing release-notes-new-features-39.md modification describing the general SmartGraph idea

3.9/graphs-general-graphs-management.md
hkernbach: added.

3.9/graphs-general-graphs-management.md
hkernbach: added.

hkernbach:
Generated examples are from branch: https://github.com/arangodb/arangodb/pull/14590
This branch also needs to be merged in parallel (right now generateSwagger.sh does not work - this needs to be fixed)

3.9/graphs-smart-graphs.md
```diff
@@ -71,6 +71,17 @@ overhead. The more subgraphs are touched the more network cost will apply.
...
+-------------------------------
+
+Hybrid SmartGraphs are capable of using Satellite Collections within their graph
+definition. Therefore, defined EdgeDefinitions between SmartCollections and
+SatelliteCollections can be created. As Satellite Collections are globally
+replicated to each participating DB-Server, graph traversals, shortest path,
+weighted traversals and k-shortest-paths queries can be executed partially
+locally on each DB-Server. This means a larger part of the query can be executed
+fully local whenever data from the Satellite part is required.
+
```
goedderz, line 84: A note that edge collections between smart- and satellite-collections are replicated like satellite collections (if I got that correctly) seems appropriate, as the amount of space needed could be unexpected otherwise.

3.9/graphs-smart-graphs.md
```diff
@@ -83,6 +94,13 @@ This ensures that graph traversals, shortest path, and k-shortest-paths queries
 can be executed locally on a DB-Server, achieving improved performance for
 these type of queries.
 
+Benefits of Hybrid Disjoint SmartGraphs
+-------------------------------
+
+Hybrid Disjoint SmartGraphs benefit of all advantages which are present in
+Hybrid SmartGraphs. The rules of Disjoint SmartGraphs, described in the chapter
+above, stay true.
+
```
goedderz, line 103: If we now add a comment how edge collections in a hybrid smart graph are replicated, we should also note here that in this case they are replicated like the disjoint smart collections (if I got that correctly), rather than like satellite collections.

3.9/release-notes-api-changes39.md
```diff
@@ -188,6 +232,11 @@ substitution.
 
 ## JavaScript API
 
+<<<<<<< HEAD
```
Simran-B, line 235: Merge conflict artifact (also below)

Simran-B (changes requested):

PR #748, opened 2021-08-13
mention extra limitations for Stream Transactions

jsteemann:
Currently we have mentioned limitations for Stream Transactions in our HTTP documentation, but not in the general "Limitations" section for transactions. 
I think the limitations should be mentioned in both places.
The PR copies the limitations paragraph from the HTTP API into the general space. This is redundant, so it is not ideal. Is there a way to write a paragraph once and use it from multiple places in the docs?


Simran-B:
As discussed with Jan earlier, there isn't a good way to include snippets of text in multiple places. Jekyll does support includes, but it doesn't work with our content structure with different "books" and versions without hacking together a custom plugin to add tags/filters that can transform links etc.

Simran-B (approved):

PR #749, opened 2021-08-18
update new features

jsteemann:


3.9/appendix-deprecated.md
```diff
@@ -23,31 +23,6 @@ considered obsolete and may get removed in a future release. They are currently
 kept for backward compatibility. There are usually better alternatives to
 replace the old features with:
 
-- **MMFiles Storage Engine**:
```
Simran-B, line 26: There are still some traces of MMFiles left, such as the `--rocksdb.exclusive-writes` option that links to this page. If we remove the section here, then the links should also be removed.

Simran-B: Will only remove from 3.9 docs. We say that users must upgrade from e.g. 3.6 to 3.7 first before they can upgrade to 3.8, so this should be fine. Gotta remove some links pointing to here though.

PR #756, opened 2021-08-30
explicitly mention subqueries in limitations

jsteemann:
As requested by @fceller in APM-77.


Simran-B (approved):
LGTM

Simran-B:
`WITH` doesn't need to be declared for dynamic access with DOCUMENT() as it seems, although there could be a corner case with mixing OneShard with sharded collections in one query and the DOCUMENT() call getting pushed to a DB-Server that cannot answer correctly. However, I couldn't find a way to get DOCUMENT() calls get pushed to DB-Servers even with both collections having a single shard and one of them referencing the other in `distributeShardsLike`.

Simran-B:
```cpp
1665     // abort on any calculations which cannot be pushed to a DB server
1666     TRI_vocbase_t& vocbase = _plan.getAst()->query().vocbase();
1667     if (type == EN::CALCULATION &&
1668         !EN::castTo<arangodb::aql::CalculationNode const*>(en)->expression()->canRunOnDBServer(vocbase.isOneShard())) {
1669       // for example, user-defined functions cannot be executed on DB servers,
1670       // as there will be no V8 engine present
1671       _sawCoordinatorCalculation = true;
1672     }
```
Line 1668 checks, whether the database is a OneShard database (i.e. whether it has the `sharding` attribute set to `"single"`). Mixing OneShard and sharded collections in a database is only possible with `"flexible"` sharding, and in that case, the OneShard optimizations do not get applied (`DOCUMENT()` etc. won't be pushed down to a DB-Server). In a `"single"` sharding database, `numberOfShards` is ignored and all collections will be OneShard.

PR #766, opened 2021-09-15
Change term vocbase to database

ronaldpetty:
Typo fix

### DIFF
--- a/3.7/data-modeling-collections-database-methods.md
+++ b/3.7/data-modeling-collections-database-methods.md
@@ -380,7 +380,7 @@ Collection Name
 <!-- arangod/V8Server/v8-vocbase.cpp -->
 
 
-selects a collection from the vocbase
+selects a collection from the database
 `db.collection-name`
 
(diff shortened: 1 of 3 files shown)


Simran-B:
That's an interesting one. Technically, it's not wrong, but the term _vocbase_ is almost exclusively used internally. VOC stands for versatile object container, and "a voc" plus three more letters is avocado - the [original name of ArangoDB](https://www.arangodb.com/2012/05/avocadodb-become-arangodb/).

I agree that it should be "database" in the end-user docs, however. To accept your patch, we need you to fill out and sign our Contributor License Agreement (CLA) for legal reasons. We use an Apache 2 CLA for ArangoDB, which can be found here: https://www.arangodb.com/documents/cla.pdf

(Or if this is too much hassle for you, then I can also close this pull request and fix it directly)

ronaldpetty:
Signed and shipped!

I feel like I have insider knowledge (voc). For the record, my family loves avocados!

PR #770, opened 2021-09-16
added MultiPolygon to supported geoJSON types

rashtao:


Simran-B (approved):
Thanks! It wasn't actually in the initial release, but added some time later.

PR #772, opened 2021-09-21
docs for APM-164

jsteemann:
https://github.com/arangodb/arangodb/pull/14796

### DIFF
--- a/3.9/http/general.md
+++ b/3.9/http/general.md
@@ -26,7 +26,7 @@ format or ArangoDB's custom [VelocyPack](https://github.com/arangodb/velocypack)
 binary format. Details on the expected format and JSON attributes can be found
 in the documentation of the individual API endpoints.
 
-Clients sending requests to ArangoDB must use either HTTP 1.0, HTTP 1.1, HTTP 2
+Clients sending requests to ArangoDB must use either HTTP 1.1, HTTP 2
 or VelocyStream. Other HTTP versions or protocols are not supported by ArangoDB.
 
 Clients are required to include the `Content-Length` HTTP header with the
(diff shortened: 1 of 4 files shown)


Simran-B (approved):

3.9/http/general.md
```diff
@@ -557,3 +553,35 @@ request forwarding as a fall-back solution.
...
+requests they send to arangod with the header `x-arango-queue-time-seconds`.
+If set, the value of the header should contain the maximum server-side
+queuing time (in seconds) that the client application is willing to accept.
+If the header is set in an incoming request, arangod will compare the current
+dequeuing time from its scheduler with the maximum queue time value contained
+in the request header. If the current queueing time exceeds the value set
+in the header, arangod will reject the request and return HTTP 412
+(precondition failed) with the error code 21004 (queue time violated).
+In a cluster, the `x-arango-queue-time-seconds` request header will be
+checked on the receiving Coordinator, before any request forwarding.
```
jsteemann, line 587:
```suggestion
In addition, client applications and drivers can optionally augment the
requests they send to arangod with the header `x-arango-queue-time-seconds`.
If set, the value of the header should contain the maximum server-side
queuing time (in seconds) that the client application is willing to accept.
If the header is set in an incoming request, arangod will compare the current
dequeuing time from its scheduler with the maximum queue time value contained
in the request header. If the current queueing time exceeds the value set
in the header, arangod will reject the request and return HTTP 412
(precondition failed) with the error code 21004 (queue time violated). 
Using a value of 0 or a non-numeric value in the header will lead to the
header value being ignored by arangod.

There is also a new metric `arangodb_scheduler_queue_time_violations_total`
that is increased whenever a request is dropped because of the requested
queue time not being satisfiable. Administrators can use this metric to monitor
overload situations.

In a cluster, the `x-arango-queue-time-seconds` request header will be
checked on the receiving Coordinator, before any request forwarding. If the
request is forwarded by the Coordinator to a different Coordinator, the 
receiving Coordinator will also check the header on its own.
Apart from that, the header will not be included in cluster-internal requests
executed by the Coordinator, e.g. when the Coordinator issues sub-requests
to DB servers or agency instances.
```

PR #784, opened 2021-10-04
Update go-getting-started.md

markburgess:
Improve the introduction for a general audience, of non-experts. This format can be a template for all the drivers. The focus is on pedagogy, to get people up and running as quickly as possible.

### DIFF
--- a/drivers/go-getting-started.md
+++ b/drivers/go-getting-started.md
@@ -228,7 +228,7 @@ type MyEdgeLink struct {
     Weight  float64 `json:"weight"`
     
 }
-'''
+```
 
 When reading data from ArangoDB with, say, ReadDocument(), 
 the API asks you to submit a variable of some type, say `MyDocumentType`, by reference
@@ -237,7 +237,7 @@ using the `&` operator:
 ```go
  var variable MyDocumentType
  mycollection.ReadDocument(nil, rawkey, &variable)
-'''
+```
 
 This submitted type is not necessarily a fixed type, but it must be a type whose members
 map (at least partially) to the named fields in the database’s JSON
--- a/drivers/go-getting-started.md
+++ b/drivers/go-getting-started.md
@@ -44,7 +44,7 @@ functions defined within it: e.g.
 driver.Function()    // package driver
 fmt.Println("hello") // package fmt
 os.Exit(0)           // package os
-'''
+```
 
 There are two Go package components needed for the Go driver.
 The main go-driver name contains an illegal character however, so we need to map
@@ -56,7 +56,7 @@ import (
               "github.com/arangodb/go-driver/http"
        driver "github.com/arangodb/go-driver”
 )
-'''
+```
 
 Now we refer to the member functions and types as: 
 
@@ -65,7 +65,7 @@ http.NewConnection(..)
 driver.NewClient(..)
 driver.Database
 driver.Collection
-'''
+```
 
 ## Configuration
 
--- a/drivers/go-getting-started.md
+++ b/drivers/go-getting-started.md
@@ -52,9 +52,8 @@ it to an alias, say “driver” for the Arango driver (or something shorter), b
 
 ```go
 import (
-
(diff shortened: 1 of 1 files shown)


drivers/go-getting-started.md
```diff
@@ -15,6 +15,58 @@ description: To use the driver, first fetch the sources into your GOPATH
...
+```go
+package.member()
+var myvariable package.typename 
+```
+
+Moreover, the package name becomes an object reference to all types and unclassified
+functions defined within it: e.g.
+
+```go
+driver.function()    // package driver
```
forgems, line 44: This is not correct. This is an unexported function and you can't call unexported package functions.

drivers/go-getting-started.md
```diff
@@ -15,6 +15,58 @@ description: To use the driver, first fetch the sources into your GOPATH
...
+Some operations (like deleting a database) cannot easily be done using
+the API, as they are too dangerous.
+
+## Go idioms
+
+Go uses package names as datatypes, and datatypes as qualifiers, to invoke 
+member functions:
+
+```go
+package.member()
```
forgems, line 36: This is not correct. This is an unexported function and you can't call unexported package functions.

drivers/go-getting-started.md
```diff
@@ -15,6 +15,58 @@ description: To use the driver, first fetch the sources into your GOPATH
...
+var myvariable package.typename
```
forgems, line 37: This is unexported type. You can't use unexported types in go. Please convert to `package.TypeName`

drivers/go-getting-started.md
```diff
@@ -31,7 +83,7 @@ running on localhost.
 import (
 	"fmt"
 
-	driver "github.com/arangodb/go-driver"
+	A "github.com/arangodb/go-driver"
```
forgems, line 86: What's the purpose of changing this alias to uppercase `A` ?

forgems: IMHO this should stay a `driver` alias or remove an alias at all (it's still going to be driver)

drivers/go-getting-started.md
```diff
@@ -43,7 +95,7 @@ conn, err := http.NewConnection(http.ConnectionConfig{
 if err != nil {
     // Handle error
 }
-c, err := driver.NewClient(driver.ClientConfig{
+client, err := A.NewClient(driver.ClientConfig{
```
forgems, line 98: IMHO changing from `driver` to `A` lowered the readability of this. UpperCase names in go should be reserved to exported functions, types, variables, and constants.

drivers/go-getting-started.md
```diff
@@ -54,12 +106,50 @@ if err != nil {
...
+	Endpoints: []string{"https://5a812333269f.arangodb.cloud:8529/"},
+})
+if err != nil {
+    // Handle error
+}
+
+// Client object
+client, err = A.NewClient(A.ClientConfig{
+	Connection: conn,
+	Authentication: A.BasicAuthentication("root", "wnbGnPpCXHwbP"),
```
forgems, line 145: I think that you should `******` instead of a real password?

drivers/go-getting-started.md
```diff
@@ -75,13 +165,239 @@ book := Book{
...
+
+// A typical vertex type must have field matching _key
+
+type MyVertexNode struct {
+
+    Key     string    `json:"_key"` // mandatory field (handle) - short name
+ 
+      // other fields … e.g.
+
+    Data    string `json: "data"`   // Longer description or bulk string data
```
forgems, line 215: The common practice is to use pointer fields for non-mandatory fields and adding `omitempty` annotation, so that you could update a single field without changing others. But I'm not sure if this is a getting-started topic.

forgems (changes requested):
Provided examples are not valid Go statements (like usage of unexported package members). Also usage of uppercased alias for package name is not idiomatic go code.

forgems (approved):
LGTM

PR #797, opened 2021-10-31
Add missing description for --server.maximal-threads default value

Simran-B:
Apparently an oversight.


santanu82021 (commented):
All missing default values are found in all versions.

PR #800, opened 2021-10-31
Add examples to DOCUMENT function for bind variables

### LINKED ISSUE #392: Usage of DOCUMENT() with bind vars
A user tried the following:
- `DOCUMENT("users/" + @_key)` - no error, but evaluates to `DOCUMENT(0) // null`
- `DOCUMENT("users/@_key")` - string literal, no bind parameter

Other possible mistakes:
- `DOCUMENT("users/", @_key)` - slash included, but this will try to find the document in a collection called `users/` and not `users` (which is not a legal name, shouldn't this raise an error?)

Solutions:
- `DOCUMENT("users", @_key)` - but note that if `@_key` is all digits, then it is necessary to wrap the value in **double-quotes** in the web UI bind parameter table! Would be helpful to not rely on color coding in the UI alone, but also display the inferred type as a badge/tag.
- `DOCUMENT(CONCAT("users/", @_key))` - possible, but unnecessary to concatenate

BTW. some of the DOCUMENT() example in the docs use collection literals (unquoted strings) instead of strings. The optimizer seems to take care of this but in earlier versions this could lead to the materialization of the whole collection.

TheJPlan:
Possibly closes #392

### DIFF
--- a/3.7/aql/functions-miscellaneous.md
+++ b/3.7/aql/functions-miscellaneous.md
@@ -221,6 +221,9 @@ DOCUMENT( users, "john" )
 
 DOCUMENT( users, [ "users/john", "users/amy" ] )
 DOCUMENT( users, [ "john", "amy" ] )
+
+DOCUMENT( @@coll, @key ) // bind vars: {"@coll": "users", "key": "john"}
+DOCUMENT( @@coll, @keys ) // bind vars: {"@coll": "users", "keys": ["john", "amy"] }
 ```
 
 `DOCUMENT(id) → doc`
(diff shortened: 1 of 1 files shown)


3.7/aql/functions-miscellaneous.md
```diff
@@ -235,6 +238,10 @@ The function can also be used with a single parameter *id* as follows:
 ```js
 DOCUMENT("users/john")
 DOCUMENT( [ "users/john", "users/amy" ] )
+
+DOCUMENT( @id ) // bind vars: {"id": "users/john"}
+DOCUMENT( @ids ) // bind vars: {"id": ["users/john", "users/amy"]}
```
cw00dw0rd, line 243: The bind var should be `ids` not `id` for the multiple users example. Otherwise, looks good to me.

ansoboleva:
Requested changes were implemented in https://github.com/arangodb/docs/pull/817.

PR #803, opened 2021-11-03
more explanations about index usage

jsteemann:


santanu82021 (commented):
All changes are reviewed and found ok.

PR #805, opened 2021-11-06
remove usages of "skiplist"

jsteemann:
in the RocksDB engine, a skiplist index is the same as a persistent
index. it is just an alias and there for compatibility reasons. we
should discourage the usage of "skiplist" indexes and instead use and
recommend "persistent" indexes.

### DIFF
--- a/3.10/administration-cluster.md
+++ b/3.10/administration-cluster.md
@@ -122,9 +122,8 @@ for a new document, but must let the server generated one automatically. This
 restriction comes from the fact that ensuring uniqueness of the primary key
 would be very inefficient if the user could specify the document key.
 
-Unique indexes (hash, skiplist, persistent) on sharded collections are
-only allowed if the fields used to determine the shard key are also
-included in the list of attribute paths for the index:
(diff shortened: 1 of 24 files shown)


ansoboleva:
@jsteemann the following pages of the drivers docs mention the skiplist index as well:

- https://www.arangodb.com/docs/devel/drivers/spring-data-reference-mapping-indexes.html
- https://www.arangodb.com/docs/devel/drivers/spring-data-reference-mapping.html
- https://www.arangodb.com/docs/devel/drivers/spring-data-reference-template-collection-manipulation.html

Should they be updated too or not?

jsteemann:
@ansoboleva : I am happy if they also get updated, but I can't say much about the drivers. I think we should check with @rashtao here. Still using the old annotations for skiplists is probably not wrong, but we should probably not advertise them or at least discourage their use for the future.

ansoboleva:
@jsteemann thanks for the explanation. @rashtao can you please help with the drivers docs here?

ansoboleva (approved):
reviewed. the drivers docs will be updated in a separate PR.

PR #811, opened 2021-11-15
Examples added to Date Functions in docs

santanu82021:


dothebart:
please note that you could rather use `EXAMPLE_AQL` so that it is actually executed as a query; see 3.9/aql/examples-combining-graph-traversals.md for example.
This would server two purposes:
- if the server moves away from the documentation, we will know
- if there is a typo in the example, we will know
- you get the actual server reply.

3.10/aql/functions-date.md
```diff
@@ -224,6 +224,25 @@ Return the weekday number of *date*.
   - 5 – Friday
   - 6 – Saturday
 
+**Examples**
+
+{% aqlexample examplevar="examplevar" type="type" query="query" bind="bind" result="result" %}
+  @startDocuBlockInline datedyofwk1
+	@EXAMPLE_AQL{datedyofwk1}
```
cw00dw0rd, line 231: My only concern is with indentation. Not sure if it matters for Docublocks but the [Readme ](https://github.com/arangodb/docs#adding-a-new-aql-example)shows keywords left aligned and states 'preferred' indentation.

ansoboleva: Looks like it doesn't matter, but for consistency and readability purposes, i'd still align pair elements, i.e. startDocuBlock/endDocuBlock and EXAMPLE_AQL/END_EXAMPLE_AQL

PR #824, opened 2021-11-25
discourage the use of arangoimp

jsteemann:
Documentation for a deprecation discussed turing team lead sync meeting today:
* deprecate _arangoimp_ alias in 3.9: arangoimp is just an alias for arangoimp
* remove _arangoimp_ in some future version, and keep _arangoimport_ around


ansoboleva (approved):
reviewed

PR #831, opened 2021-12-02
license documentation after final pm decision

kvahed:
Updated license documentation to reflect the final PM decision:

* Time after installation without enforcement (currently 3 days): 1 hour
* Time before license expiration with daily warnings (currently 14 days): 7 days
* Time after expiration switch to Read-Only (currently 14 days): 0 days


ansoboleva (approved):
reviewed

PR #850, opened 2021-12-16
BV set to have valid JSON in examples.

santanu82021:


dothebart:
imho the docublock names should be a bit more speaking than `docN` rather like `FUNCTION_DOCUMENT_N` so they can be more easily be distinguished and located in the text like when using grep.

ansoboleva:
Duplicate of #851

PR #861, opened 2022-01-05
Added information about lookahead index hint.

maierlars:
Docs PR for https://github.com/arangodb/arangodb/pull/14745


3.10/indexing-multi-dim.md
```diff
@@ -126,6 +126,18 @@ FOR app IN appointments
...
+## Lookahead Index Hint
+
+Using the lookahead index hint can increase the performance for certain use cases. Specifiying a lookahead value greater than zero makes the index fetch more documents, that are no longer in the search box, before seeking to the next lookup position. 
+Because the seek operation is computationally expensive, probing more documents before seeking may reduce the number of seeks, if matching documents are found. Please keep in mind that it might also affect performance negatively.
+
+You can specify the lookahead value using the OPTIONS keyword:
+```js
+FOR app IN appointments OPTIONS {lookahead: 32}
+    FILTER f <= app.to
+    FILTER app.from <= t
```
jsteemann, line 138: @maierlars : shouldn't the parameters be exactly opposite, i.e. 
```suggestion
    FILTER @to <= app.to
    FILTER app.from <= @from
```
?

3.10/indexing-multi-dim.md
```diff
@@ -126,6 +126,18 @@ FOR app IN appointments
     RETURN app
 ```
 
+## Lookahead Index Hint
+
+Using the lookahead index hint can increase the performance for certain use cases. Specifiying a lookahead value greater than zero makes the index fetch more documents, that are no longer in the search box, before seeking to the next lookup position.
```
jsteemann, line 131:
```suggestion
Using the lookahead index hint can increase the performance for certain use cases. Specifiying a lookahead value greater than zero makes the index fetch more documents that are no longer in the search box, before seeking to the next lookup position. 
```

3.10/indexing-multi-dim.md
```diff
@@ -126,6 +126,18 @@ FOR app IN appointments
     RETURN app
 ```
 
+## Lookahead Index Hint
+
+Using the lookahead index hint can increase the performance for certain use cases. Specifiying a lookahead value greater than zero makes the index fetch more documents, that are no longer in the search box, before seeking to the next lookup position. 
+Because the seek operation is computationally expensive, probing more documents before seeking may reduce the number of seeks, if matching documents are found. Please keep in mind that it might also affect performance negatively.
```
jsteemann, line 132:
```suggestion
Because the seek operation is computationally expensive, probing more documents before seeking may reduce the number of seeks, if matching documents are found. Please keep in mind that it might also affect performance negatively if documents are fetched unnecessarily.
```

3.10/indexing-multi-dim.md
maierlars: It depends on what you want to query. So what do you had in mind?

jsteemann: At least it doesn't look very intuitive to me. My assumption was that `f <= t`, but obviously you can also query with `f > t`. 
Anyway, I think we should at least replace `f` and `t` with bind parameter names, because it is else unclear what `f` and `t` mean here.

jsteemann (approved):
LGTM

PR #862, opened 2022-01-05
Thread default value updated in docs

santanu82021:
Old: 
The number of threads can be controlled via the --threads option, which defaults to 2.

Change to: 
The number of threads can be controlled via the --threads option. The default value changed from 2 to the maximum of 2 and the number of available CPU cores.


ansoboleva (approved):
reviewed

PR #864, opened 2022-01-05
add more indexHints: `disableIndex` and `maxProjections`

jsteemann:
**Depends on 3.9.1**
Docs PR for https://github.com/arangodb/arangodb/pull/15446 (devel) and https://github.com/arangodb/arangodb/pull/15470 (3.9.1)


Simran-B:
Should update the `reduce-extraction-to-projection` description to mention the new options:
https://www.arangodb.com/docs/stable/aql/execution-and-performance-optimizer.html#list-of-optimizer-rules

jsteemann:
@Simran-B : I think the new options should be documented primarily for the AQL `FOR` operation, which I think they are via this PR.
IMHO, optimizer rules shouldn't be documented in detail, because they come and go and often change under the hood. As I can see from the link you posted, the optimizer rule documentation is pretty out of date (still mentions RocksDB storage engine, as if there would be another storage engine available....). Anyway, I will fix it in this case, but in general optimizer rule documentation should not go into detail. The exact details for each optimizer rule can be _very_ tricky, so it will be a massive endeavor to document all that and even more to keep the docs up to date.

jsteemann:
@Simran-B : done.

ansoboleva (approved):
reviewed

PR #865, opened 2022-01-06
slightly improve documentation for vector functions

jsteemann:
* added version number in which functions were introduced
* mention that the functions can return `null` and produce warnings


alexbakharew (approved):
LGTM

PR #868, opened 2022-01-11
micro improvements

jsteemann:
Add AF deprecation notice to `release-notes-upgrading-changes39.md` as well, to increase its visibility.
The deprecation notice was previously only in `release-notes-new-features39.md`, where it does not belong IMO.

The PR also contains some other micro changes.


ansoboleva (approved):
reviewed

PR #874, opened 2022-01-18
OAS-3817 | Oasis Disk Auto-Sizing Feature

Simran-B:
- Add info about disk auto-sizing to deployments page - [preview](https://deploy-preview-874--zealous-morse-14392b.netlify.app/docs/3.8/oasis/deployments.html#oneshard)
- Update oasisctl docs to v1.12

Should we explicitly state that an increase in disk size increases the costs?

Cloudflare cache will need to be cleared for the updated images to show for all users.


ewoutp (approved):
LGTM

PR #876, opened 2022-01-20
Fix small inaccuracies about threads

neunhoef:
A customer noted that we have some small inconsistencies between
documentation and implementation. This fixes them.

This is around number of server threads.


jsteemann (approved):
LGTM!

Simran-B (approved):
Slightly improved wording, undid change to 3.6 content as that version is EoL.

PR #877, opened 2022-01-20
BTS-719 | Query explain output for 3.9 traversal optimizations

Simran-B:
Adds examples of how two new optimizations can be observed in the query explain output as requested in https://arangodb.atlassian.net/browse/BTS-719


ansoboleva (approved):
reviewed

PR #888, opened 2022-01-31
Change some default values for RocksDB memory usage

jsteemann:
https://github.com/arangodb/arangodb/pull/15627


ansoboleva (approved):
reviewed

PR #902, opened 2022-02-08
note library version upgrades

jsteemann:
All of the updated library versions are in devel aready, so we can as well update the docs.


ansoboleva (approved):
reviewed

PR #911, opened 2022-02-14
BTS-386 | HTTP Error Handling and Response Codes

Simran-B:
1. Unsure if we should state both the actual behavior and the previously documented behavior or just the former ("closes connection, not return NNN" vs. "closes connection")

2. My understanding is that 504 Gateway Timeout isn't currently returned to users and that there's the idea to return 503 in cases where requests can safely be retried and 504 if it's uncertain but that the server currently replies with 503 in all cases while requests may not be retried safely. We should only document the current behavior, therefore adding 503 with a remark about idempotency.

3. Unclear whether Content-Size > 1 GB **and** an actual payload of that size lead to 413 Payload Too Large. Nothing happened in my test, both the client and the server were seemingly idle, no error, no connection closure

Behavior tested using 3.8.5.1. Other versions to be tested.

### DIFF
--- a/3.10/http/general.md
+++ b/3.10/http/general.md
@@ -346,7 +346,8 @@ HTTP layer:
 - client requests using an HTTP version signature different than *HTTP/1.0* or
   *HTTP/1.1* will get an *HTTP 505* (HTTP version not supported) error in return.
 - ArangoDB will reject client requests with a negative value in the
-  `Content-Length` request header with *HTTP 411* (Length Required).
+  `Content-Length` request header by closing the connection, not with
+  *HTTP 411* (Length Required).
 - ArangoDB doesn't support POST with `Transfer-Encoding: chunked` which forbids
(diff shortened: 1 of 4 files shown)


neunhoef:
I do not think we should document the old behaviour. Let's await what backporting is done and decide then.

About the 504: I think we should make it our policy to return 503 only if it is guaranteed that no harm is done (or for idempotent requests). But this is not yet done, and probably not even decided yet. And then we need an error code to say that we do not know if anything has happened, and that would probably be 504. So maybe we have to wait with the documentation a bit longer until we have sorted this one out.

If a `ContentLength` of less than 1GB is given, then we will just read as many bytes and then interpret the request with what we have. Any excess data will then be interpreted as the header of the next request, with whatever consequences this might have.

neunhoef (approved):
LGTM

PR #913, opened 2022-02-17
Oasis: Update permissions, pre-defined roles, and oasisctl docs

Simran-B:
`oasisctl list roles` lists `ipwhitelist-admin` - wasn't this role renamed to `ipallowlist-admin`?


ewoutp (approved):
LGTM

PR #918, opened 2022-02-21
optionally use range delete markers in WAL for truncates

jsteemann:
Docs PR for https://github.com/arangodb/arangodb/pull/15791
Enable truncate WAL markers in the cluster too. This can speed up truncate operations for large collections/shards.
The newest version of arangosync can now handle truncate markers, so it is safe to activate truncate markers on the arangod side as well.
There is a feature toggle `--rocksdb.use-range-delete-in-wal` to turn off the usage of truncate markers in case of emergency.


ansoboleva (approved):
reviewed

PR #925, opened 2022-02-24
Add back content to persistent index

Simran-B:
Some commented out information about hash and skiplist indexes is still relevant for the persistent index.

The indexing chapter needs more refactoring (DOC-38) but this PR removes the HTML comments from the markup, at least.

Moved "Vertex-centric indexes" below "Edge Index" because it makes more sense and to have all sections about different index types above the sections with general remarks and options.

Removed mentions of in-memory indexes. Unclear whether the following statement is still true with the primary index not being an in-memory index anymore:

> A persistent index does not store pointers into the primary index, but instead it stores a document's primary key. To retrieve a document via a persistent index with an index value lookup, there will therefore be an additional O(1) lookup into the primary index to fetch the actual document.

-> Removed statement because it's not true anymore (both index types use internal document IDs). Alternative statements would potentially be confusing because it's a complicated matter, see https://arangodb.slack.com/archives/C1PJT16QP/p1645700419169119


3.10/indexing-index-basics.md
```diff
@@ -94,84 +93,81 @@ user-defined indexes.
...
-Non-unique hash indexes have an amortized complexity of O(1) for insert, update, and
-removal operations. That means non-unique hash indexes can be used on attributes with 
-low cardinality. 
+If we for example have a persistent index on the attributes `_from` and 
+`timestamp` of an edge collection, we can answer the above question
+very quickly with a single range lookup in the index.
 
-If a hash index is created on an attribute that is missing in all or many of the documents,
-the behavior is as follows:
+You can create sorted persistent indexes that index the special edge attributes
```
jsteemann, line 121:
```suggestion
You can create sorted persistent indexes that indexes the special edge attributes
```

3.10/indexing-index-basics.md
```diff
@@ -94,84 +93,81 @@ user-defined indexes.
...
+in arangosh. Then, queries like
 
+```js
+FOR v, e, p IN 1..1 OUTBOUND "V/1" edges
+  FILTER e.timestamp >= "2018-07-09"
+  RETURN p
+```
 
+will be considerably faster in case there are many edges originating
+from vertex `"V/1"` but only few with a recent time stamp. Note that the
```
jsteemann, line 142:
```suggestion
from vertex `"V/1"` but only few with a recent timestamp. Note that the
```

3.10/indexing-index-basics.md
```diff
@@ -212,94 +208,54 @@ sort step:
...
-but instead it stores a document's primary key. To retrieve a document via a persistent
-index via an index value lookup, there will therefore be an additional O(1) lookup into 
-the primary index to fetch the actual document.
-
-As the persistent index is sorted, it can be used for point lookups, range queries and sorting
-operations, but only if either all index attributes are provided in a query, or if a leftmost 
-prefix of the index attributes is specified.
-
-<!-- TODO
+If a persistent index is created on an attribute that is missing in all or many
```
jsteemann, line 250: Covered in the sections before already. I suggest removing this to shorten the section!

3.10/indexing-index-basics.md
```diff
@@ -599,59 +556,31 @@ only if the query filters on the indexed attribute using the `IN` operator. The
...
-Since ArangoDB 3.0 one can create sorted indexes (type
-"persistent") that index the special edge attributes `_from` or `_to`
-and additionally other attributes. Since ArangoDB 3.1, these are used
-in graph traversals, when appropriate `FILTER` statements are found
-by the optimizer.
-
-For example, to create a vertex centric index of the above type, you 
-would simply do
+Ensure uniqueness of relations in edge collections
+--------------------------------------------------
```
jsteemann, line 560: "Ensure ..." sounds a bit like a command.
```suggestion
Ensuring uniqueness of relations in edge collections
----------------------------------------------------
```

3.10/indexing-index-basics.md
Simran-B: "index the special edge attributes" (plural) seems correct?

3.10/indexing-index-basics.md
```diff
@@ -599,59 +556,31 @@ only if the query filters on the indexed attribute using the `IN` operator. The
...
 
-```js
-db.edges.ensureIndex({"type":"persistent", "fields": ["_from", "timestamp"]});
-```
+You can create a combined index over the edge attributes `_from` and `_to`
+with the unique option enabled to prevent duplicate relations from being created.
```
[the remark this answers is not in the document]
Simran-B: It's a specific use case (graph) that comes up often, so we added a dedicated section that we can point users to.

3.10/indexing-index-basics.md
```diff
@@ -599,59 +556,31 @@ only if the query filters on the indexed attribute using the `IN` operator. The
...
 
-in arangosh. Then, queries like
+For example, a document collection *verts* might contain vertices with the document
+handles `verts/A`, `verts/B` and `verts/C`. Relations between these documents can
+be stored in an edge collection *edges* for instance. Now, you may want to make sure
+that the vertex `verts/A` is never linked to `verts/B` by an edge more than once.
+This can be achieved by adding a unique, non-sparse persistent index for the
+fields `_from` and `_to`:
```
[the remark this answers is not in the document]
Simran-B: It's a specific use case (graph) that comes up often, so we added a dedicated section that we can point users to.

ansoboleva (approved):
reviewed

PR #930, opened 2022-03-03
Added Spark Datasource to drivers list

rashtao:
- added ArangoDB Datasource for Apache Spark to the drivers list
- updated the url of python-arango


ansoboleva (approved):
reviewed

PR #936, opened 2022-03-10
BTS-820 | Adjust JWT timeout

Simran-B:
3.9+ have a default token lifetime of 1 hour whereas previous versions expired after 1 month. Can be controlled by the `--server.session-timeout` startup option.


ansoboleva (approved):
reviewed

PR #941, opened 2022-03-12
AQL: k Paths recognizes pseudo-keyword TO

Simran-B:
I also noticed that SEARCH is listed as high-level operation and we state that all of them would be reserved keywords, but this is contradicted by the list of non-reserved pseudo-keyword that includes SEARCH. Shall I add a remark?


ansoboleva (approved):
reviewed

PR #946, opened 2022-03-16
icu-language option

alexbakharew:
Added documentation to icu-language option

### DIFF
--- a/3.10/programs-arangod-general.md
+++ b/3.10/programs-arangod-general.md
@@ -40,29 +40,34 @@ parameter pid-file is given, then the server will report an error and exit.
 
 ## Default Language
 
-`--default-language default-language`
+`--default-language language-name`
 
 The default language is used for sorting and comparing strings. The language
 value is a two-letter language code (ISO-639) or it is composed by a two-letter
-language code with and a two letter country code (ISO-3166). Valid languages are
-"de", "en", "en_US" or "en_UK".
+language code followed by a two letter country code (ISO-3166). For example:
+"de", "en", "en_US", "en_UK".
 
 The default default-language is set to be the system locale on that platform.
 
 ## ICU Language
 
-`--icu-language icu-language`
+<small>Introduced in: v3.9.1</small>
 
-The ICU language is also used for sorting and comparing strings. But in this case language 
-is used without any changes in ICU settings. Using this option it is possible to get absolutely correct sorting and comparing order. The language value is a two-letter language code (ISO-639) or it is composed by a two-letter
-language code with and a two letter country code (ISO-3166). Valid languages are
-"de", "en", "en_US" or "en_UK".
+`--icu-language language-name`
 
-If there are defined `--default-language` and `--icu-language` it will leads to server crash at start.
+The ICU language is also used for sorting and comparing strings. With this option however,
+you can get the sorting and comparing order exactly as it's defined in the ICU standard. 
+The language value can be a two-letter language code (ISO-639), a two-letter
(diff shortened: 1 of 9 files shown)


3.9/programs-arangod-general.md
```diff
@@ -42,13 +42,26 @@ parameter pid-file is given, then the server will report an error and exit.
...
+<small>Introduced in: v3.9.1</small>
+
+`--icu-language icu-language`
+
+The ICU language is also used for sorting and comparing strings. But in this case language 
+is used without any changes in ICU settings. Using this option it is possible to get absolutely correct sorting and comparing order. The language value is a two-letter language code (ISO-639) or it is composed by a two-letter
+language code with and a two letter country code (ISO-3166). Valid languages are
+"de", "en", "en_US" or "en_UK".
+
+If there are defined `--default-language` and `--icu-language` it will leads to server crash at start.
```
Simran-B, line 63: crash -> raise a fatal error?

ansoboleva:
Docs for https://github.com/arangodb/arangodb/pull/15912

3.9/programs-arangod-general.md
alexbakharew: Yes

3.10/programs-arangod-general.md
```diff
@@ -42,13 +42,27 @@ parameter pid-file is given, then the server will report an error and exit.
...
 "de", "en", "en_US" or "en_UK".
 
 The default default-language is set to be the system locale on that platform.
 
+## ICU Language
+
+`--icu-language icu-language`
+
+The ICU language is also used for sorting and comparing strings. But in this case language 
+is used without any changes in ICU settings. Using this option it is possible to get absolutely correct sorting and comparing order. The language value is a two-letter language code (ISO-639) or it is composed by a two-letter
```
Dronplane, line 57:
```suggestion
is used without any changes in ICU settings. Using this option it is possible to get pure ICU defined sorting and comparing order. The language value is a two-letter language code (ISO-639) or it is composed by a two-letter
```

ansoboleva (approved):
reviewed

Dronplane (approved):
LGTM

3.10/programs-arangod-general.md
```diff
@@ -40,15 +40,34 @@ parameter pid-file is given, then the server will report an error and exit.
...
-language code with and a two letter country code (ISO-3166). Valid languages are
-"de", "en", "en_US" or "en_UK".
+language code followed by a two letter country code (ISO-3166). For example:
+"de", "en", "en_US", "en_UK".
 
 The default default-language is set to be the system locale on that platform.
 
+## ICU Language
+
+<small>Introduced in: v3.9.1</small>
```
KVS85, line 54: Probably not needed here once mentioned in 3.9 docs.

KVS85 (commented):
Please, see review comment(s).

3.10/programs-arangod-general.md
Simran-B: It's useful at the very least because it's not in the .0 release (covered by the release notes). One could argue that the manually written startup option pages don't require version hints because we have this information in the generated startup option tables but I'm in favor of having it here, too, for convenience.

Ideally, there would only be a single place instead of two for every startup option information but that requires a design solution first and likely a change to the tooling second.

KVS85:
- [ ] https://jenkins.arangodb.biz:3456/view/Documentation/job/arangodb-ANY-examples/1783/

PR #949, opened 2022-03-18
[DE-190] Spark: edge schema validation

rashtao:
The schema of a Dataframe being written to an an edge collection must have:
- a non nullable string field named `_from`
- a non nullable string field named `_to`


ansoboleva (approved):
reviewed

PR #955, opened 2022-03-21
Allow more examples to be generated on Windows

Simran-B:
- Make file paths OS-agnostic
- `fs` is available by default in example generation but not regular arangosh. A proper example should include `var fs = require("fs");` because that's what a user would need to run, but it actually unsets the global fs variable (because of JS weirdness?), leading to errors. Rename the variable name in the tooling? Or not polute the global scope in general?
  https://github.com/arangodb/arangodb/blob/devel/Documentation/Scripts/exampleHeader.js
  Every accidental name collision may cause trouble. Helper functions do need to be available, however. Unsure whether side-effects are a necessity or if `eval()` could be sandboxed with an IIFE...
  Do any existing examples need a `require("fs")` if we remove `fs` from the global scope?
- Hot backup-related examples can't be fixed as they fail because the feature isn't available on Windows

### DIFF
--- a/3.10/analyzers.md
+++ b/3.10/analyzers.md
@@ -964,12 +964,13 @@ Create and use a `classification` Analyzer with a stored "cooking" classifier to
 @EXAMPLE_ARANGOSH_RUN{ClassificationAnalyzerModelSetup}
 var fs = require("fs");
 var internal = require("internal");
+var tmpPath = fs.getTempPath();
 try {
-    fs.makeDirectory("/tmp/embeddingsModels");
+    fs.makeDirectory(fs.join(tmpPath, "embeddingsModels"));
 } catch (e) {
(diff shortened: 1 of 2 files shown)


dothebart (approved):
nice!

Simran-B:
Somewhat related: https://github.com/arangodb/arangodb/pull/16778

Simran-B:
Analyzer examples cannot be generated, see #1063.
The only other known use of file paths is for a debug dump example (absolute path) and #1414 (relative path).
There is no visible use of the `fs` module in any example, only in invisible asserts.

`fs.getTempPath()` returns a different path in every arangosh process, which is a potential issue for the new tooling.
Closing, since the new tooling will run in Docker anyway.

PR #878, opened 2022-01-20
AQL String Functions: Describe caseInsensitive parameter, improve examples and port to @EXAMPLE_AQL

Simran-B:
- `caseInsensitive` parameter not sometimes missing, sometimes in the wrong place.
- Examples ported to `@EXAMPLE_AQL` (3.10 only)
- Added a few more examples, renamed a couple to match convention, removed some that didn't provide additional value. Left example queries involving a View unchanged.
- Modified some descriptions and examples that didn't match actual behavior (likely changed around 3.0 noticed)
- Factored out regular expression syntax description and moved it to the end of the page

https://jenkins.arangodb.biz:3456/view/Documentation/job/arangodb-ANY-examples/1605/

PR #958, opened 2022-03-28
Feature/aql-keep-recursive

Simran-B:
- Add documentation for the new AQL document function `KEEP_RECURSIVE()`
  https://github.com/arangodb/arangodb/pull/15923
- Convert examples in `functions-document.md` to @EXAMPLE_AQL but only in 3.10
- Improve some of the other function descriptions and examples, as well as the formatting following
  #878

https://jenkins.arangodb.biz:3456/view/Documentation/job/arangodb-ANY-examples/1781/

Preview: https://deploy-preview-958--zealous-morse-14392b.netlify.app/docs/3.10/aql/functions-document.html


ansoboleva (approved):
reviewed

3.10/aql/functions-document.md
```diff
@@ -345,142 +622,365 @@ individual parts as separate attributes.
...
-```
+**Examples**
+
+Recursively remove `baz` attributes, by passing an array with the attribute key:
+
+    {% aqlexample examplevar="examplevar" type="type" query="query" bind="bind" result="result" %}
+    @startDocuBlockInline aqlUnsetRecursive_7
+    @EXAMPLE_AQL{aqlUnsetRecursive_7}
+    LET doc = { foo: { bar: { foo: 1, baz: 2 }, baz: 3 }, baz: 4 }
+    RETURN UNSET_RECURSIVE(doc, ["baz"])
```
KVS85, line 920: Looks like a meaningless example comparing to original `UNSET_RECURSIVE( doc, [ "_id", "_key", "foo", "bar" ] )`.

KVS85 (commented):
LGTM.

3.10/aql/functions-document.md
Simran-B: The KEEP_RECURSIVE and UNSET_RECURSIVE examples use the same example object and I don't really want to use an object that has nested, system attribute-like keys, which would be the only way to make it meaningful for the different query variations, while still using the original arguments for the function call here.

The point of the specific example is to show that you can pass an array. For the keep counterpart it made sense to show passing multiple keys but here it seemed better to let some attributes remain in the object.

I agree that many of the examples are contrived but making them more real-world would also make them less succinct.

PR #960, opened 2022-03-28
TTL index expects Unix timestamp in seconds

Simran-B:
Corrects a statement that timestamps would be expected in milliseconds. Found by @solisoft


ansoboleva (approved):
reviewed

PR #981, opened 2022-04-21
Report true referencing DocuBlock when missing a struct

Simran-B:
Fixes the error messages like the following:
https://app.netlify.com/sites/zealous-morse-14392b/deploys/6261204442c805000846bbb8#L121-L178

The shared structs are missing for 3.9, will re-generate them.


Simran-B:
Also addressed a minor rendering issue because of a missing new line before **Examples**:

image

image

PR #986, opened 2022-04-26
adds arangosearch mention to indexes page

cw00dw0rd:
- The ArangoSearch section isn't mentioned on the indexing page but I think it should be since it lives there now.

### DIFF
--- a/3.10/indexing.md
+++ b/3.10/indexing.md
@@ -19,3 +19,4 @@ There are special sections for
   - [Multi-dimensional Indexes](indexing-multi-dim.html)
   - [Geo-spatial Indexes](indexing-geo.html)
   - [Vertex-centric Indexes](indexing-vertex-centric.html)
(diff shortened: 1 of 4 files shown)


Simran-B:
Yes, thanks!
I would word it differently, though. Views are more than inverted indexes, and I would like to avoid any confusion with the upcoming per-collection inverted indexes.

cw00dw0rd:
I agree that there is a better way to phrase it. My goal was to inform the reader about why ArangoSearch is being included in the indexing section, since it is more than just an index. 
Have any suggestions?

Since the guides cover all of ArangoSearch we could just do something like
```
How to use ArangoSearch views, analyzers, and functions.
```

ansoboleva:
Implemented @cw00dw0rd's suggestion.

ansoboleva (approved):
reviewed

PR #988, opened 2022-04-27
[DE-210] Spark ignoreNullFields

rashtao:
Added option:
- `ignoreNullFields`: whether to ignore null fields during serialization, `false` by default


ansoboleva (approved):
reviewed

PR #989, opened 2022-04-28
Getting Started Restructure

Simran-B:
Summary:
- New **Introduction** with a focus on graphs
  - Removed documentation overview (books/parts)
  - Removed old feature list
  - Removed community support section (out of place)
- New **Use Cases** page `[WIP]`
  - ArangoDB as Graph / Document / Key-Value Database, Search Engine, for Machine Learning
- New pages about **Features**
  - Overview with a comparison of Community and Enterprise Edition and Cloud vs. On-Prem
  - Detailed feature lists for CE and EE (separate pages because they are pretty long)
  - Made _Highlights_ page a sub-chapter
- New **Quick Start Guide** `[WIP]`
  - Selection of interactive tutorials
  - ArangoDB in the cloud (moved Oasis _Getting Started_ and _Overview_ to Manual)
  - On-premise / local installation
  - How to interact with ArangoDB (web interface, arangosh, drivers & integrations, REST API), some explanation of key concepts but may get moved to _Data Model & Concepts_
  - Coming from SQL (couldn't find a better place for this existing content)
- Updated **Data Model & Concepts** `[WIP]`
  - Better explanation of key concepts, reduced content redundancy (but still more to refactor)

PR #1005, opened 2022-05-26
Update tutorial-join.md

jeengbe:
Incorrect example


Simran-B:
Thank you for bringing this up!
The intention is to show a query that is as simple as possible. Therefore, I think it's better to adjust the example result rather than the query. Will be fixed by https://github.com/arangodb/docs/pull/989/commits/faa97f3663872da0c97e1f3e3a967a5d2c4c4033

PR #1007, opened 2022-05-31
add several configuration options for Pregel jobs

jsteemann:
docs PR for https://github.com/arangodb/arangodb/pull/16332

### DIFF
--- a/3.10/programs-arangod-pregel.md
+++ b/3.10/programs-arangod-pregel.md
@@ -5,15 +5,15 @@ layout: default
 
 ## Pregel job parallelism
 
-Pregel jobs have configurable minimum, maximum and default parallelism values.
-These parallelism options can be used by administrators to set concurrency defaults and bounds 
+Pregel jobs have configurable minimum, maximum, and default parallelism values.
+Administrators can use these parallelism options to set concurrency defaults and bounds 
 for Pregel jobs on an instance level. Each individual Pregel job can set its own parallelism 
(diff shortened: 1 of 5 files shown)


3.10/programs-arangod-pregel.md
```diff
@@ -0,0 +1,111 @@
...
+---
+# ArangoDB Server Pregel Options
+
+## Pregel job parallelism
+
+Pregel jobs have configurable minimum, maximum and default parallelism values.
+These parallelism options can be used by administrators to set concurrency defaults and bounds 
+for Pregel jobs on an instance level. Each individual Pregel job can set its own parallelism 
+value using the job's `parallelism` option, but the job's effective parallelism will be clamped 
+to the bounds defined by `--pregel,min-parallelism` and `--pregel.max-parallelism`.
```
markuspf, line 12:
```suggestion
to the bounds defined by `--pregel.min-parallelism` and `--pregel.max-parallelism`. 
```

markuspf (approved):
LGTM (mod typos) as general documentation;

At some point we might need to write sizing-guidelines or put more detail in places.

For example it is clear to me that the max/min parallelism settings limit each individual pregel job; this does not limit how many cores might be used by the pregel subsystem as a whole.

Also memory mapped files do not cause any overhead as long as data still fits in ram, because nothing is actually written to disk, but as soon as data does go to disk this has performance implications. (the extent of which we cannot give right now because we don't know)

ansoboleva (approved):
reviewed

PR #1012, opened 2022-06-06
correction for drop graph, collection handling

cw00dw0rd:
The wording of comment indicated supplying `true` results in the collections being preserved but the opposite is true. It might make more sense to update the comment but wasn't sure certain on the intent of the example.

### DIFF
--- a/3.10/graphs-satellite-graphs-management.md
+++ b/3.10/graphs-satellite-graphs-management.md
@@ -208,9 +208,9 @@ can `truncate` the graph if you just want to get rid of the data.
         var graph_module = require("@arangodb/satellite-graph");
         var relation = graph_module._relation("edges", "vertices", "vertices");
         var graph = graph_module._create("myGraph", [relation], []);
-        graph._deleteEdgeDefinition("edges");
-        graph._removeVertexCollection("vertices");
-        graph_module._drop("myGraph", false); // does not drop any collections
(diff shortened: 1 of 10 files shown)


Simran-B:
This is indeed confusing. I think the point was to show that, even if you tell `_drop()` to delete the underlying collections, they will still be around if you remove them from the graph definition prior to dropping the graph. In other words, the example de-couples the collections from the graph, then drops it, but because the collections aren't part of the named graph anymore, they will not be deleted despite the `true` argument.

```js
graph._deleteEdgeDefinition("edges");
graph._removeVertexCollection("vertices");
graph_module._drop("myGraph", true); // does not drop any collections
```

The example makes sense, assuming that it wants to show the particular case of collections being removed from the graph definition before dropping the graph, but it could use a better description. How about the following?

```js
graph._deleteEdgeDefinition("edges");      // remove edge collection from graph definition
graph._removeVertexCollection("vertices"); // remove vertex collection from graph definition
graph_module._drop("myGraph", true);       // does not drop any collections because there are none left in the graph definition
```

cw00dw0rd:
Thank you for the clarification and I see now the original example was true just a little confusing. I have added your suggestion as I think that helps clarify it.

Simran-B (approved):
Removed changes to EoL versions from the PR, fixed indentation, and slightly reworded the descriptions.

PR #1029, opened 2022-06-18
Update Reduce Memory Footprint tutorial

Simran-B:
- 3.9.2 lowers the minimum `--cache.size` to 0 to disable it
- 3.10.0 adds optional hash caches for persistent indexes, whose size is also controlled by the startup option

### DIFF
--- a/3.10/tutorials-reduce-memory-footprint.md
+++ b/3.10/tutorials-reduce-memory-footprint.md
@@ -116,17 +116,20 @@ Also see:
 - [RocksDB Server Options](programs-arangod-options.html#rocksdb)
 - [Write Buffer Manager](https://github.com/facebook/rocksdb/wiki/Write-Buffer-Manager){:target="_blank"}
 
-Edge-Cache
-----------
+Index Caches
+------------
 
 ```
---cache.size 10485760
+--cache.size 0
 ```
 
-This option limits the ArangoDB edge [cache](programs-arangod-cache.html) to 10
-MB. If you do not have a graph use-case and do not use edge collections, it is
-possible to use the minimum without a performance impact. In general, this
-should correspond to the size of the hot-set of edges.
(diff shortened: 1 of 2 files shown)


3.9/tutorials-reduce-memory-footprint.md
```diff
@@ -116,17 +116,19 @@ Also see:
...
---cache.size 10485760
+--cache.size 0
 ```
 
-This option limits the ArangoDB edge [cache](programs-arangod-cache.html) to 10
-MB. If you do not have a graph use-case and do not use edge collections, it is
-possible to use the minimum without a performance impact. In general, this
-should correspond to the size of the hot-set of edges.
+This option disables the ArangoDB edge [cache](programs-arangod-cache.html).
+In versions before v3.9.2, you can limit the size to a minimum of `10485760` (10 MB).
```
jsteemann, line 127:
```suggestion
In versions before v3.9.2, you can limit the size to a minimum of `1048576` (1 MB).
```

3.10/tutorials-reduce-memory-footprint.md
```diff
@@ -116,17 +116,20 @@ Also see:
...
---cache.size 10485760
+--cache.size 0
 ```
 
-This option limits the ArangoDB edge [cache](programs-arangod-cache.html) to 10
-MB. If you do not have a graph use-case and do not use edge collections, it is
-possible to use the minimum without a performance impact. In general, this
-should correspond to the size of the hot-set of edges.
+This option disables the in-memory index [caches](programs-arangod-cache.html).
+In versions before v3.9.2, you can limit the size to a minimum of `10485760` (10 MB).
```
jsteemann, line 127:
```suggestion
In versions before v3.9.2, you can limit the size to a minimum of `1048576` (1 MB).
```

PR #1031, opened 2022-06-20
note/homebrew install not available for M1

nerpaula:
added note: homebrew install is currently not available for M1 (Apple silicon)


dothebart:
since the current approach for these users is, to run the docker container without AVX, should this be noted as a workaround here? The docker container name is `arangodb/arangodb:3.9.X-noavx`

nerpaula:
@dothebart afaik, this is a development workaround that is not officially supported. if so, I don't think it should be added here.

3.9/installation-mac-osx.md
```diff
@@ -12,6 +12,12 @@ You can install ArangoDB on macOS in different ways:
 - [_DMG_ Package](#package-installation)
 - [_tar.gz_ Archive](#installing-using-the-archive)
 
+{% hint 'warning' %}
+Currently, the ArangoDB installation on macOS via Homebrew,
```
Simran-B, line 16: "currently" should be avoided: https://developers.google.com/style/timeless-documentation

We can be explicit and say that 3.9 does not run on Apple's ARM chips. 3.8 and older do work via Rosetta 2 emulation, and 3.10 will have native ARM support.

ansoboleva (approved):
lgtm, i wouldn't mention any not official workarounds here

3.10/release-notes-new-features310.md
```diff
@@ -9,6 +9,13 @@ The following list shows in detail which features have been added or improved in
 ArangoDB 3.10. ArangoDB 3.10 also contains several bug fixes that are not listed
 here.
 
+Native ARM Support
+------------------
+
+ArangoDB is now available for the ARM architecture, in addition to the x86-64
+architecture. It can run on Apple silicon (e.g. M1 chips) without emulation.
+It also runs on other ARM chips under Linux and Windows for ARM.
```
Simran-B, line 17: Will we support Windows for ARM? Or only Linux and macOS?
What is the minimum ARM architecture that we will support? ARMv8 (because of SIMD/NEON)?

joerg84 (approved):

3.10/release-notes-new-features310.md
KVS85: @Simran-B AFAIK no Windows support is planned. Yes, it will be ARMv8 at this point and NEON is used.

KVS85 (changes requested):
See comments.

3.10/release-notes-new-features310.md
```diff
@@ -13,8 +13,16 @@ Native ARM Support
 ------------------
 
 ArangoDB is now available for the ARM architecture, in addition to the x86-64
-architecture. It can run on Apple silicon (e.g. M1 chips) without emulation.
-It also runs on other ARM chips under Linux and Windows for ARM.
+architecture.
+
+It can natively run on Apple silicon (e.g. M1 chips). It was already possible to
+run 3.8.x and older versions on these systems via Rosetta 2 emulation, but not
+3.9.x because of its use of AVX instructions, which Rosetta does not emulate.
```
KVS85, line 20: 3.9.x because of its use of AVX instructions, which Rosetta **2** does not emulate.

3.10/release-notes-new-features310.md
```diff
@@ -13,8 +13,16 @@ Native ARM Support
...
+3.10.x runs on this hardware again, but now without emulation.
+
+ArangoDB 3.10.x also runs on 64-bit ARM (AArch64) chips under Linux.
+The minimum requirement is an ARMv8 chip with SIMD support in the form of the
```
KVS85, line 24: The minimum requirement is an ARMv8 chip with Neon (SIMD) support in the form of the

KVS85 (changes requested):
Maybe it makes sense to emphasize that till 3.10 any possibilities to run on ARM were non-production unsupported cases.

PR #1033, opened 2022-06-20
Dotnet driver

tjoubert:
Updates to dotnet driver docs.


drivers/dotnet-getting-started.md
```diff
@@ -3,115 +3,16 @@ layout: default
...
-### Replace a document
-
-```csharp
-// Fully update document
-item.Description = "Some item with some more description";
-await adb.Document.PutDocumentAsync(
-    $"MyCollection/{item._key}",
-    item);
-```
+Browse the full reference for [ArangoDBNetStandard](dotnet-reference/Home.md).
```
Simran-B, line 18: Where is this supposed to point to?

tjoubert: Hi @Simran-B , thanks for your feedback. There is a reference documentation of the driver that is generated by DocFx. The idea was to create a subfolder dotnet-reference where Home.md would be the main page of that reference documentation. I deleted the reference documentation because it was too big (over 2000 md pages). I will remove the reference to it here for now and we will put it back in once we decide on where to place the reference documentation.

drivers/dotnet-installation.md
```diff
@@ -0,0 +1,14 @@
+---
+layout: default
+---
+# Installation
```
Simran-B, line 4: This doesn't actually show how to install the driver. It is okay in the case of Nuget I suppose because the referenced site shows different ways to obtain the driver but as a beginner, I wouldn't know how to continue.

As there is very little content here and on the Getting Started page, I would suggest merging it.

drivers/dotnet-getting-started.md
Simran-B: Generated reference documentation should be hosted through GitHub Pages, with the option to fully automate the generation and deployment. For reference, see how [arangojs](https://github.com/arangodb/arangojs) does it (hosted site: https://arangodb.github.io/arangojs/).

tjoubert: We're working on pushing the reference to Github Pages. Once that's done, I will update the link here to point to it.

drivers/dotnet-installation.md
tjoubert: @Simran-B, I have added more details to the installation page. I was thinking of keeping Getting Started as an index page with links to others providing more details about each topic. Do you still think we should merge it?

drivers/dotnet-getting-started.md
tjoubert: Hi @Simran-B , I have successfully published the reference docs and added it's link to the Getting Started page. Kindly review. Thanks

drivers/dotnet-installation.md
tjoubert: @Simran-B, let me know your final thoughts on this. Thanks

ansoboleva (approved):
reviewed

PR #1045, opened 2022-06-27
Jekyll version plugin: Try a fallback value when comparing versions

Simran-B:
Potential workaround for Netlify build error
`` _plugins/versions/version.rb:45:in `<=>': undefined method `version' for nil:NilClass (NoMethodError) ``

Additionally, print debug info in the hope that it helps to identify the circumstances under which the error occurs (it only does sporadically, and I can't recall this happening in local builds). Note that the following is printed all the time, including local builds:

```
Version method from_path called with invalid path '/404.html'
Version method from_path called with invalid path '/'
Version method from_path called with invalid path '/redirects.json'
```

https://app.netlify.com/sites/zealous-morse-14392b/deploys/62b97f3e7372ca0009611e28#L120-L122

This does not fail builds, but additional messages like these are noteworthy.


ansoboleva:
@Simran-B i don't quite understand the purpose of this PR. Also, wouldn't invest time and effort into fixing something that's not critical and is about to be replaced anyhow.

ansoboleva (approved):

Simran-B:
Error caught for a 3.7 <=> nil comparison: https://github.com/arangodb/docs/pull/1109#issuecomment-1245924825

PR #1060, opened 2022-07-14
DOC-145: Added documentation for APM-84

cpjulia:
Documentation for https://github.com/arangodb/arangodb/pull/16421.


3.10/aql/invocation-with-arangosh.md
```diff
@@ -242,6 +242,44 @@ In the ArangoDB Enterprise Edition there is an additional parameter:
...
+
+- `spillOverThresholdNumRows`: This parameter allows for input data and 
+  intermediate results to be spilled onto disk for the execution of a query 
+  after the memory usage reaches the value in bytes this parameter holds. This 
+  is used for decreasing the memory usage during the query execution. This 
+  parameter is experimental and is only taken into account if a path for a 
+  directory where the temporary data would be stored is provided with the 
+  server startup option 
+  `--temp.intermediate-results-path`. 
+  Default value: 128MB.
```
nerpaula, line 280: both parameters having the same name, is that correct?

cpjulia: no, it's `spillOverThresholdMemoryUsage`, fixed in latest commit

3.10/aql/invocation-with-arangosh.md
```diff
@@ -242,6 +242,44 @@ In the ArangoDB Enterprise Edition there is an additional parameter:
   into sync. The default value is 60.0 (seconds). When the max time has been reached
   the query will be stopped.
 
+Additional parameters for spilling data from the query onto disk
```
Simran-B, line 245: None of the other options have their own headline. I wonder whether we should group some of the others as well, or not do that for any of them. Giving each option an own headline would be another option and allow to link to each single one (we can hide them in the ToC if necessary)

3.10/aql/invocation-with-arangosh.md
```diff
@@ -242,6 +242,44 @@ In the ArangoDB Enterprise Edition there is an additional parameter:
...
+-----------------------------------------------------------------
+
+  There are two additional parameters that would allow for spilling 
+  intermediate data from the query onto disk for descreasing the memory usage.
+  Note: The option of spilling data from RAM onto disk is experimental and off
```
Simran-B, line 250: Use hint box

3.10/release-notes-new-features310.md
```diff
@@ -427,6 +427,53 @@ arangoexport \
...
 ```
 
+
+Query changes for decreasing memory usage
+-----------------------------------------
+
+With the parameters mentioned below, queries can execute with storing input
+and intermediate results temporarily on disk for decreasing of memory usage
+when it reaches a certain threshold.
+Note: this feature is experimental and off by default. The query results will
```
Simran-B, line 437: Remarks should only be in the feature documentation, not the release notes. The option descriptions should probably be removed (but new start options should still need to be mentioned).

3.10/aql/invocation-with-arangosh.md
cpjulia: Actually, the options above this are explained one by one starting in line 141, so I added these extra ones too

ansoboleva (approved):
reviewed

PR #1059, opened 2022-07-14
DOC-123 | Document computed attributes APM-199

Simran-B:
Initial documentation for the computed values feature.

TBD in a separate PR:
- Release notes
- Examples
- HTTP API

https://deploy-preview-1059--zealous-morse-14392b.netlify.app/docs/3.10/data-modeling-documents-computed-values.html

PR #1065, opened 2022-07-19
DOC-123 | Document computed attributes APM-199 (part 2)

Simran-B:
Follow-up to #1059 

Upstream: https://github.com/arangodb/arangodb/pull/16584 and https://github.com/arangodb/arangodb/pull/16600

- [x] `overwrite` option is still called `override`, unclear whether it will be changed - renamed via https://github.com/arangodb/arangodb/pull/16600
- [x] Computed values **can** be based on one another **but** not others that run on the same `computeOn` event (or can they, with unpredictable results?), e.g. `update` will see the values `insert` wrote. Julia to confirm that the observed behavior is correct. - The feature works slightly different than described in the DD but seems to work as expected. `keepNull` works less like in AQL Analyzers and more like in UPDATE queries. Adjusted the docs.
- [x] Invalid expression cannot be saved but `RETURN ASSERT(false, 'error')` will produce a runtime error no matter the value of `failOnWarning` (it's only for turning warnings into errors) - It's now mentioned that invalid expressions are rejected immediately. Should we additionally say that AQL errors can still be thrown at runtime? (There is nothing special about Computed Values in this regard compared to full AQL)

Preview: https://deploy-preview-1065--zealous-morse-14392b.netlify.app/docs/3.10/data-modeling-documents-computed-values.html

Needs regenerated 3.10 examples.


3.10/data-modeling-documents-computed-values.md
```diff
@@ -28,11 +28,20 @@ every document, or to automatically combine multiple attributes into one,
 possibly with filtering and a conversion to lowercase characters, to then index
 the attribute and use it to perform case-insensitive searches.
 
+## Using Computed Values
+
+Computed values are defined per collection using the `computedFields` property,
```
jsteemann, line 33:
```suggestion
Computed values are defined per collection using the `computedValues` property,
```

3.10/data-modeling-documents-computed-values.md
```diff
@@ -28,11 +28,20 @@ every document, or to automatically combine multiple attributes into one,
...
+Computed value definitions are included in dumps, and the attributes they added,
+too, but no expressions are executed when restoring dumps. The collections and
+documents are restored as they are in the dump and no attributes are recalculated.
+
 ## JavaScript API
 
-Computed values are defined per collection, either when creating the collection
-or by modifying the collection later on. The collection property is called
-`computedFields` and accepts an array of objects.
+The `computedFields` collection property accepts an array of objects.
```
jsteemann, line 44:
```suggestion
The `computedValues` collection property accepts an array of objects.
```

3.10/data-modeling-documents-computed-values.md
```diff
@@ -76,14 +92,14 @@ You can access the document data via the `@doc` bind variable. It contains the
...
-
-Computed value expressions have the following requirements:
+document to work with, including the user-provided values.
 
-- The expression must start with a `RETURN` statement.
+Computed value expressions have the following properties:
 
-- No `FOR` loops, `LET` statements, and subqueries are allowed in the expression.
-  `FOR` loops can be substituted using the [array expansion operator `[*]`](aql/advanced-array-operators.html#inline-expressions),
+- The expression must start with a `RETURN` operation and cannot contains any
```
jsteemann, line 99:
```suggestion
- The expression must start with a `RETURN` operation and cannot contain any
```

3.10/data-modeling-documents-computed-values.md
```diff
@@ -93,9 +109,15 @@ Computed value expressions have the following requirements:
...
-  like any access of non-existent attributes.
+- You cannot access the result of another computed value that is generated on
+  the same `computeOn` event.
+  
+  For example, two computed values that are generated on `insert` can not see
+  the result of the other. Referencing the attributes results in an implicit
+  `null` value. Computed values that are generated on `update` or `replace` can
+  see the results of the previous `insert` computations, however. They cannot
+  see the new values of other `update` and `replace` computations, regardless of
+  the order of the computed value definitions in the `computedFields` property.
```
jsteemann, line 120:
```suggestion
  the order of the computed value definitions in the `computedValues` property.
```

3.10/data-modeling-documents-computed-values.md
```diff
@@ -28,11 +28,20 @@ every document, or to automatically combine multiple attributes into one,
...
+documents are restored as they are in the dump and no attributes are recalculated.
+
 ## JavaScript API
 
-Computed values are defined per collection, either when creating the collection
-or by modifying the collection later on. The collection property is called
-`computedFields` and accepts an array of objects.
+The `computedFields` collection property accepts an array of objects.
 
 `db._create(<collection-name>, { computedFields: [ { … }, … ] })`
```
jsteemann, line 46:
```suggestion
`db._create(<collection-name>, { computedValues: [ { … }, … ] })`
```

ansoboleva (approved):
reviewed

PR #1066, opened 2022-07-19
correct collection name

diyarfaraj:
I believe it should be "users" at the end of the query.

### DIFF
--- a/3.9/aql/operations-update.md
+++ b/3.9/aql/operations-update.md
@@ -79,7 +79,7 @@ FOR i IN 1..1000
 ```js
 FOR u IN users
   FILTER u.active == false
-  UPDATE u WITH { status: 'inactive' } IN backup
+  UPDATE u WITH { status: 'inactive' } IN users
 ```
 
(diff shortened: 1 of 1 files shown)


Simran-B:
The example is correct, it shows that you can read from a collection using a FOR loop and write to a different collection. Is the example description not clear enough perhaps?

> An update operation may update arbitrary documents which do not need to be identical
to the ones produced by a preceding `FOR` statement

diyarfaraj:
> The example is correct, it shows that you can read from a collection using a FOR loop and write to a different collection. Is the example description not clear enough perhaps?
> 
> > An update operation may update arbitrary documents which do not need to be identical
> > to the ones produced by a preceding `FOR` statement

For me with a few months experience with Arango, that specific explanation was not clear, to be honest. But maybe it's just me :)

Simran-B:
Let me know if this is clear enough: https://github.com/arangodb/docs/pull/1074

diyarfaraj:
very nice, much better thanks!

PR #1069, opened 2022-07-21
DOC-125 | Add docs for All Shortest Paths AQL graph traversal

Simran-B:
3.10 examples need to be re-generated.


aMahanna (approved):
thank you :)

PR #1074, opened 2022-07-25
AQL: Clarify that UPDATE/REPLACE can target a different collection than what a FOR loop read from

Simran-B:
Supersedes #1066


ansoboleva (approved):
reviewed

PR #1075, opened 2022-07-27
Fix generation of example for aql analyzer

alexbakharew:
- [ ] https://jenkins.arangodb.biz/view/Documentation/job/arangodb-ANY-examples/2225/


KVS85 (approved):
LGTM.

3.10/analyzers.md
```diff
@@ -755,9 +755,9 @@ Otherwise the position is set to the respective array index, 0 for `"A"`,
     |   "FOR d IN SPLIT(@param, '-') RETURN d"
       }, ["frequency", "norm", "position"]);
       var coll = db._create("coll");
+      var doc = db.coll.save({ text: "A-B-C-D" });
```
Simran-B, line 758: What difference does this change make?

alexbakharew: If we will create view and insert doc into collection, we may observe that query (which we run immediately after creation) return nothing. And examples in docs will be incorrect. 

It may be connected with fact that we should wait a little bit until doc will be indexed

Simran-B: The example contains a (hidden) query to wait until the View is updated:

`db._query("FOR d IN view OPTIONS { waitForSync: true } LIMIT 1 RETURN true");`

Therefore, I don't think that saving the doc before creating the view makes any difference

Simran-B: Summarizing the subsequent Slack conversation:
`waitForSync` doesn't appear to guarantee that the View is updated if documents are added after the View creation, but according to Alexey and Vadim it works (reliably) if the documents are added before the View creation. So this change is needed.
@ansoboleva please go ahead with this PR.

Simran-B (approved):

PR #1048, opened 2022-06-30
Read from followers: APM-296

neunhoef:
This adds documentation for the read-from-followers feature.

- Add x-arango-allow-dirty-read header to read-only operations.
- Add snippets from this PR:

PR #1076, opened 2022-07-31
DOC-151 | Reading from followers in clusters is EE only (APM-296)

Simran-B:
Follow-up PR to #1048


Simran-B:
Netlify failed again, and also in Julia's PR. Let's merge #1045 and see if it helps to avoid these build errors.

ansoboleva (approved):
reviewed

PR #1077, opened 2022-08-02
Clarify fastText Model Type

geenen124:
The classification and nearest_neighbors analyzers only accept supervised models.


hkernbach (changes requested):
Good change :) 
We should also mention that this is an Enterprise-Only Feature.

hkernbach (approved):

ansoboleva (approved):
reviewed

PR #1082, opened 2022-08-08
DOC-166 | Add `computedValues` collection property to more JS API pages

Simran-B:
Also adds a description for `schema` that was missing. I haven't checked if more properties are missing.

The important changes are in this commit: https://github.com/arangodb/docs/pull/1082/commits/a42066dededd681dd32381a822681fbe12336e99


alexbakharew (approved):
LGTM regarding Computed Values

ansoboleva (approved):
lgtm

PR #1092, opened 2022-08-18
DOC-140 | Locality-sensitive hashing / MinHash

Simran-B:
Essential docs for LSH / MinHash, mostly reference documentation and release notes.
In-depth content to be added later when we have an example dataset.

Also covers DOC-163 and DOC-169

Needs 3.10 and devel example generation


3.10/analyzers.md
```diff
@@ -976,11 +993,55 @@ Create different `segmentation` Analyzers to show the behavior of the different
...
+    @startDocuBlockInline analyzerMinHash
+    @EXAMPLE_ARANGOSH_OUTPUT{analyzerMinHash}
+      var analyzers = require("@arangodb/analyzers");
+      var analyzerMinHash = analyzers.save("minhash5", "minhash", { analyzer: { type: "segmentation", properties: { break: "alpha", case: "lower" } }, numHashes: 5 }, ["frequency", "norm", "position"]);
+      var analyzerSegment = analyzers.save("segment", "segmentation", { break: "alpha", case: "lower" }, ["frequency", "norm", "position"]);
+    | db._query(`
+    |   LET str1 = "The quick brown fox jumps over the lazy dog."
+    |   LET str2 = "The fox jumps over the crazy dog."
+    |   RETURN {
+    |     approx: JACCARD(TOKENS(str1, "minhash5"), TOKENS(str2, "minhash5")),
```
Simran-B, line 1030: @gnusi Is this a meaningful example?

3.10/aql/functions-miscellaneous.md
```diff
@@ -446,6 +446,74 @@ guaranteed to remain the same in future versions of ArangoDB. The hash values
 should therefore be used only for temporary calculations, e.g. to compare if two
 documents are the same, or for grouping values in queries.
 
+### MINHASH()
+
+`MINHASH(values, numHashes) → hashes`
```
gnusi, line 451: Would be nice to provide an example of how it can be used together with stored values.

gnusi, line 451: This must be an array functions

gnusi (changes requested):

3.10/aql/functions-miscellaneous.md
Simran-B: Based on the data type of the first parameter? In that case, we would need to move a lot of functions to different pages, e.g. MEDIAN(). The functions are rather grouped by "topic", and the misc function page already lists hash functions, so that seemed like the best place to me.

Simran-B: Good suggestion, but that should not be in the AQL reference. The idea is to get the 3.10 reference docs done first, and then we can add additional content like guides (for MinHash, I would like to add an example page under ArangoSearch).

3.10/release-notes-api-changes310.md
```diff
@@ -91,6 +91,13 @@ following attributes:
 - `keepNull` (boolean, _optional_, default: `true`)
 - `failOnWarning` (boolean, _optional_, default: `false`)
 
+#### MinHash Analyzer
+
+The `/_api/analyzer` endpoint supports a new [`minhash` Analyzer](analyzers.html#minhash)
+type. It has two properties, `analyzer` (object) and `numHashes` (number).
```
Simran-B, line 97:
```suggestion
The `/_api/analyzer` endpoint supports a new [`minhash` Analyzer](analyzers.html#minhash)
type in the Enterprise Edition. It has two properties, `analyzer` (object) and `numHashes` (number).
```

3.10/analyzers.md
Simran-B: Fine according to Andrei

ansoboleva (approved):
reviewed

PR #1096, opened 2022-08-23
DOC-132 | Nested search with ArangoSearch Views

Simran-B:
Basic description of nested search, limited to Views for now.


Dronplane (commented):
We state " Nested search (Enterprise Edition)" only in new-features but I suggest  to mention EE in all places to avoid confusion.

3.10/arangosearch-nested-search.md
```diff
@@ -0,0 +1,218 @@
...
+
+If you leave out the quantifier, it defaults to `ANY`. The conditions need to be
+fulfilled by at least one sub-object, but more than one sub-object may meet the
+conditions. With a quantity of `1`, it would need to be one match exactly.
+Similarly, ranges require an exact match between the minimum and maximum number,
+including the specified boundaries. To require two or more sub-objects to
+fulfill the conditions, you can use `AT LEAST (2)`, and so on.
+
+{% hint 'info' %}
+The `ALL` quantifier of the Question mark operator is not supported in View queries.
```
gnusi, line 136: by `View` here we mean `arangosearch` views, but this is supported for `search-alias` views. In general I believe we have to revisit documentation so we can make a distinction between `arangosearch` and `search-alias` explicit.

3.10/arangosearch-nested-search.md
```diff
@@ -0,0 +1,218 @@
...
+
+Using the question mark operator without quantifier and filter conditions (`[?]`)
+cannot utilize View indexes.
```
gnusi, line 139: It seems that was fixed by @Dronplane

3.10/arangosearch-nested-search.md
```diff
@@ -0,0 +1,218 @@
...
+attributes, and the built-in `text_en` is used for the `comments`.
+
+A possible query is to search for frames with damaged corners that are not wider
+than 80, using a question mark operator to check the `part` and `comments`, and
+a nested question mark operator to check the `type` and `value`:
+
+```aql
+FOR doc IN viewName
+  SEARCH doc.dimensions[? FILTER CURRENT.part == "frame" AND
+         ANALYZER(TOKENS("corner damage", "text_en") ALL == CURRENT.comments, "text_en") AND
```
gnusi, line 209: As for inverted indexes and `search-alias` views we discourage using `ANALYZER` function since actual field value is inferred from a field properties automatically, I suggest we either leave a comment or refactor an example.

3.10/aql/operations-search.md
```diff
@@ -156,6 +156,22 @@ attribute is an array. `IN` and `==` as part of array comparison operators are
...
+```aql
+FOR doc IN myView
+  SEARCH doc.dimensions[? FILTER CURRENT.type == "height" AND CURRENT.value > 40]
+  RETURN doc
+```
+
+It allows you to match nested objects in arrays that satisfy multiple conditions
+each, and optionally define how often these conditions should be fulfilled for
+the entire array. You need to configure the View specifically for this type of
+search using the `nested` [link property](../arangosearch-views.html#link-properties).
```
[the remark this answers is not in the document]
Simran-B: to be done in the inverted index PR

3.10/arangosearch-nested-search.md
Simran-B: Good point. Should we call them ArangoSearch Views and Search Alias Views, or rather something like Views of type `arangosearch` / `search-alias`?

3.10/arangosearch-nested-search.md
Simran-B: My understanding is that it is supposed to work but expected not to use the View index but fall back to a post-filter operation, hence the statement "cannot utilize View indexes". I can make it clearer, though:
```suggestion
is possible but cannot utilize View indexes.
```

3.10/arangosearch-nested-search.md
Simran-B: Nested search using inverted indexes needs to be covered in a separate PR, on a different page, but the `search-alias` case needs to be addressed here, indeed. Should there be separate sections about the different View types but on the same page? Or only remarks about using the ANALYZER() function in combination with search-alias Views? That would be on every example page, however...

3.10/aql/operations-search.md
```diff
@@ -156,6 +156,22 @@ attribute is an array. `IN` and `==` as part of array comparison operators are
...
+
+```aql
+FOR doc IN myView
+  SEARCH doc.dimensions[? FILTER CURRENT.type == "height" AND CURRENT.value > 40]
+  RETURN doc
+```
+
+It allows you to match nested objects in arrays that satisfy multiple conditions
+each, and optionally define how often these conditions should be fulfilled for
+the entire array. You need to configure the View specifically for this type of
```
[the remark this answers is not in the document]
Simran-B: to be done in the inverted index PR

3.10/aql/advanced-array-operators.md
```diff
@@ -303,4 +302,49 @@ The above will return:
...
+`ANY`. The following quantifiers are supported:
+
+- Integer numbers for exact quantities (e.g. `2`)
+- Number ranges for a quantity between the two values (e.g. `2..3`)
+- `NONE` (equivalent to `0`)
+- `ANY`
+- `ALL`
+- `AT LEAST`
+
+The quantifier needs be followed by a `FILTER` operation if you want to specify
```
jsteemann, line 333:
```suggestion
The quantifier needs to be followed by a `FILTER` operation if you want to specify
```

3.10/arangosearch-nested-search.md
```diff
@@ -0,0 +1,221 @@
...
+            "value": {}
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+Note the `nested` property instead of a `fields` property, configuring the View
```
jsteemann, line 110: Is `nested` available in the Enterprise Edition only?

Simran-B: To be precise, you can use the question mark operator in the community edition and it will give you the expected results, but it won't get any index acceleration, because that is an enterprise feature. I haven't checked whether the nested property is rejected/ignored in CE or if it's accepted and persisted, but it's definitely useless.

3.10/aql/advanced-array-operators.md
```diff
@@ -303,4 +302,49 @@ The above will return:
...
+
+| Question mark operator | Inline filter with length check |
+|:-----------------------|:--------------------------------|
+| `arr[? <number> FILTER <conditions>]` | `LENGTH(arr[* FILTER <conditions>]) == <number>`
+| `arr[? <min>..<max> FILTER <conditions>]` | `IN_RANGE(LENGTH(arr[* FILTER <conditions>]), <min>, <max>, true, true)`
+| `arr[? NONE FILTER <conditions>]` | `LENGTH(arr[* FILTER <conditions>]) == 0`
+| `arr[? ANY FILTER <conditions>]`  | `LENGTH(arr[* FILTER <conditions>]) > 0`
+| `arr[? ALL FILTER <conditions>]`  | `LENGTH(arr[* FILTER <conditions>]) == LENGTH(arr)`
+| `arr[? AT LEAST (<number>) FILTER <conditions>]` | `LENGTH(arr[* FILTER <conditions>]) >= <number>`
+| `arr[?]` | `LENGTH(arr) > 0`
```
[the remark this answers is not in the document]
Simran-B: Should every equivalent expression have an IS_ARRAY() condition?

3.10/release-notes-new-features310.md
```diff
@@ -148,6 +148,57 @@ db._query(`FOR doc IN food_view
 See [Search highlighting with ArangoSearch](arangosearch-search-highlighting.html)
 for details.
 
+This feature is available in all editions but nested search queries are
```
jsteemann, line 151:
```suggestion
This feature is available in all editions, but nested search queries are
```

3.10/release-notes-new-features310.md
```diff
@@ -148,6 +148,57 @@ db._query(`FOR doc IN food_view
...
+
+```aql
+FOR doc IN viewName
+  SEARCH doc.dimensions[? FILTER CURRENT.type == "height" AND CURRENT.value > 40]
+  RETURN doc
+```
+
+See [Nested search in ArangoSearch](arangosearch-nested-search.html) for details.
+
 This feature is only available in the Enterprise Edition.
```
jsteemann, line 202: The above text says it is available in all editions (not just enterprise), but optimized only in enterprise.

3.11/aql/advanced-array-operators.md
```diff
@@ -303,4 +302,49 @@ The above will return:
...
+`ANY`. The following quantifiers are supported:
+
+- Integer numbers for exact quantities (e.g. `2`)
+- Number ranges for a quantity between the two values (e.g. `2..3`)
+- `NONE` (equivalent to `0`)
+- `ANY`
+- `ALL`
+- `AT LEAST`
+
+The quantifier needs be followed by a `FILTER` operation if you want to specify
```
jsteemann, line 333:
```suggestion
The quantifier needs to be followed by a `FILTER` operation if you want to specify
```

3.11/arangosearch-nested-search.md
```diff
@@ -0,0 +1,221 @@
...
+            "value": {}
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+Note the `nested` property instead of a `fields` property, configuring the View
```
jsteemann, line 110:
```suggestion
Note the usage of a `nested` property instead of a `fields` property, configuring the View
```

Simran-B:
@joerg84 Please clarify how to word the docs regarding Nested search support in Community vs. Enterprise Edition.

The question mark operator is available in all Editions, which means you **can** perform nested searches to some extent, but without index acceleration and there are some caveats:
- `FOR doc IN coll FILTER doc.arr[? FILTER ...]` is a kind of nested search and works in CE (but can be slow)
- `FOR doc IN view FILTER doc.arr[? FILTER ...]` is possible (can be slow) but the nested attributes need to be indexed using `fields` or `includeAllFields` in the View definition, otherwise, it will not find anything.
  UX issue: `nested` is silently ignored in CE, so you may be under the belief that the nested attributes were indexed but are in fact not
- `FOR doc IN view SEARCH doc.arr[? FILTER ...]` throws an error in CE (not implemented) and works in EE.
  UX issue: If you index the nested attributes with `fields` instead of `nested` in EE, it can **never** match something using such a query and View definition and should potentially warn the user about it.

3.10/aql/advanced-array-operators.md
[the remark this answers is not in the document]
Simran-B: Expanding non-arrays, which is syntactically only allowed for object literals and indirectly via a variable, results in `[]`. This means that in most cases there is an indirect check whether it is an array or not. There are some corner cases (if the LENGTH check includes 0) where it would give false positive results without IS_ARRAY() but that seems rather esoteric.

3.10/arangosearch-nested-search.md
Simran-B: In the PR for the inverted index, we will remove ANALYZER() if possible and otherwise have two examples, for a Search Alias View and a traditional one (many functions accept an optional analyzer argument, e.g. PHRASE)

3.10/arangosearch-nested-search.md
Simran-B: Jörg to clarify how to word the docs regarding the different editions.

PR #1098, opened 2022-08-25
DOC-176/add EnterpriseGraphs to Release Notes 3.10

nerpaula:


3.10/release-notes-new-features310.md
```diff
@@ -459,6 +459,22 @@ created in a cluster.
...
 This feature is only available in the Enterprise Edition.
 
+## EnterpriseGraphs (Enterprise Edition)
+
+The 3.10 version of ArangoDB introduces a specialized version of SmartGraphs, 
+called EnterpriseGraphs. 
+
+EnterpriseGraphs come with a concept of automated sharding key selection,
+meaning that the sharding key is randomly selected while ensuring that all
+vertices with the same sharding key and their adjacent edges are co-located
```
mchacki, line 469:
```suggestion
meaning that the sharding key is randomly selected while ensuring that all
 their adjacent edges are co-located
```

mchacki (changes requested):
Minor correction.

For Enterprise Graphs there cannot be two vertices with the same sharding key. (only because of collisions, but nothing anyone can control)

PR #1103, opened 2022-08-30
DOC-179/arangosync-update

nerpaula:
- updated limitations, added ARM64 support
- added note for document size limitations

### DIFF
--- a/3.7/troubleshooting-dc2-dc.md
+++ b/3.7/troubleshooting-dc2-dc.md
@@ -133,6 +133,11 @@ default of `1h`.
 
 ## What to do in case of a document that exceeds the message queue limits
 
+{% hint 'info' %}
+Starting with ArangoSync version 2.0.0, there is no limitation in terms of
+document size anymore.
+{% endhint %}
+
 If you insert/update a document in a collection and the size of that document
 is larger than the maximum message size of your message queue, the collection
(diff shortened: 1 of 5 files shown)


Simran-B:
FYI: Different vendors and people use different names for the chip/instruction set architectures.

While "AMD64" is not wrong, it can be misleading because it isn't limited to AMD CPUs, and Unix-based operating systems report this architecture as "x86-64" (or "x86_64").

With ARM, there are even more subtleties between the different names. The 64-bit ARM instruction set is called A64 and the execution state is called "AArch64". Apple calls the platform "arm64" in their compiler, but other compilers use "aarch64".

For the [download snippets](https://github.com/arangodb/oskar/tree/master/snippets/3.10) on the website, I went with "x86-64 instruction set" and "64-bit ARM chips (AArch64)". While this may not be entirely correct either, we should probably be consistent.

3.7/troubleshooting-dc2-dc.md
```diff
@@ -133,6 +133,11 @@ default of `1h`.
 
 ## What to do in case of a document that exceeds the message queue limits
 
+{% hint 'info' %}
```
informalict, line 136: Should not we add this info in 3.8 and 3.9 ?

informalict (approved):
LGTM with a little comment

PR #1108, opened 2022-09-07
Dockerfile base image change

dandimeo:
Changed base dockerfile image from ubuntu:bionic to ruby:2.6 to avoid ruby build version issues


Simran-B:
The latest 2.6 ruby is apparently 2.6.10 - should we go for that?

PR #1110, opened 2022-09-10
Update go-example-requests.md

whoiscarlo:
I changed `c, err := driver.NewClient(driver.ClientConfig{`   to  `client, err := driver.NewClient(driver.ClientConfig{`. So that in the next block titled **Opening a database**, it would make sense where the variable `client` came from. It's a minor change but it makes everything consistent.

### DIFF
--- a/drivers/go-example-requests.md
+++ b/drivers/go-example-requests.md
@@ -13,7 +13,7 @@ conn, err := http.NewConnection(http.ConnectionConfig{
 if err != nil {
     // Handle error
 }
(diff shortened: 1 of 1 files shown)


Simran-B:
Fixed three occurrences of the incorrect variable name in all currently supported versions with in commit https://github.com/arangodb/docs/commit/8a43acad96656066a0967a8bda61576e07682c73

whoiscarlo:
Hey @Simran-B sorry for the delay in response. I never saw this notification. I just sent that signed the pdf and sent the email over. Thank you again

PR #1115, opened 2022-09-14
document some RocksDB changes

jsteemann:
Document some internal changes to RocksDB for version 3.10, which have been missing before.


ansoboleva (approved):
reviewed

PR #1148, opened 2022-10-21
Change default output format of arangoexport from `json` to `jsonl`

jsteemann:


ansoboleva:
@Simran-B please follow up on this PR according to https://arangodb.atlassian.net/browse/DOC-218

Simran-B (approved):

ansoboleva (changes requested):
@Simran-B Please also update this page:

programs-arangoexport.md - it should also mention JSONL as it's the default output format now.

Simran-B:
@ansoboleva Do you refer to "like JSON, CSV or XML"? It's a non-exhaustive list. Should XGMML also be mentioned?

ansoboleva:
@Simran-B yes, this bit. The list doesn't have to be exhaustive, but it should mention the default format.

Simran-B:
JSONL is less known than JSON, so we should mention both, but the only format not mentioned would then be XGMML. BTW. the proper name is JSON Lines, not JSONL. Which one should we use?

ansoboleva (approved):

PR #1149, opened 2022-10-21
DOC-216 | ArangoSearch v3.10.1

Simran-B:
Updates 3.10 and 3.11 examples but without user-visible changes.

### DIFF
--- a/3.10/aql/functions-arangosearch.md
+++ b/3.10/aql/functions-arangosearch.md
@@ -252,6 +252,7 @@ specified data type.
   - `"null"`
   - `"bool"` / `"boolean"`
   - `"numeric"`
+  - `"type"` (matches `null`, `boolean`, and `numeric` values)
   - `"string"`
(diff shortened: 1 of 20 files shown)


3.10/aql/functions-arangosearch.md
```diff
@@ -277,20 +279,74 @@ by the specified **analyzer**.
...
+  an error if used outside of a [`SEARCH` operation](operations-search.html) or
+  a `FILTER` operation that uses an inverted index.
 
 ```aql
 FOR doc IN viewName
   SEARCH EXISTS(doc.text, "analyzer", "text_en")
   RETURN doc
 ```
 
+#### Testing for nested fields
```
Dronplane, line 291: BTW function EXISTS from the very beginning do support "type" keyword
like `EXISTS(d.foo, "type")`. Is it intentional to not document this?

Dronplane (approved):
LGTM

PR #1153, opened 2022-10-25
Mention new `shardingStrategy` value in release notes

Simran-B:
Vertex collections of EnterpriseGraphs have the shardingStrategy set to the value "enterprise-hex-smart-vertex" - which was added to the docs in #1144, except the API changes page


hkernbach (approved):
LGTM, thank you.

nerpaula (approved):

PR #1167, opened 2022-10-31
Restructure Architecture chapter

nerpaula:


3.10/architecture.md
```diff
@@ -1,11 +1,49 @@
 ---
 layout: default
+description: >-
+  An overview of ArangoDB's deployment modes
```
Simran-B, line 4: The deployment modes are just one aspect of the architecture. I actually wonder if we should describe them in a separate chapter like this at all or rather on each respective page about how to deploy ArangoDB using a particular mode, and have an additional guide page to select the right deployment mode for the use case (which is a different type of content than a description of the architecture).

ansoboleva (approved):

PR #1181, opened 2022-11-10
DOC-215 | Clarify difference between UPDATE/REPLACE syntaxes

Simran-B:
Preview:
- https://deploy-preview-1181--zealous-morse-14392b.netlify.app/docs/3.10/aql/operations-update.html
- https://deploy-preview-1181--zealous-morse-14392b.netlify.app/docs/3.10/aql/operations-replace.html


3.10/aql/operations-update.md
```diff
@@ -25,32 +25,72 @@ UPDATE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></cod
...
+The following query adds or updates the `name` attribute of the document
+identified by the key `my_key` in the `users` collection. The key is passed as
+a string in the `keyExpression`. The attributes to add or update are passed
+separately as the `document` object:
+
+```aql
+UPDATE "my_key" WITH { name: "Jon" } IN users
+```
+
+If the `document` object may contain a `_key` attribute but it is ignored.
```
romanatarango, line 86: An extra "If".

romanatarango, line 86:
```suggestion
The `document` object may contain a `_key` attribute, but it is ignored.
```

3.10/aql/operations-update.md
```diff
@@ -69,20 +109,32 @@ FOR u IN users
...
-  UPDATE CONCAT('test', i) WITH { foobar: true } IN users
+  UPDATE CONCAT('test', i) WITH { name: "Paula" } IN users
 ```
 
-The documents it modifies can be in a different collection than
-the ones produced by a preceding `FOR` operation:
+Target a different collection
+-----------------------------
+
+The documents an `UPDATE` operations modifies can be in a different collection
```
romanatarango, line 131:
```suggestion
The documents an `UPDATE` operation modifies can be in a different collection
```

3.10/aql/operations-update.md
```diff
@@ -167,64 +219,62 @@ update non-existing documents or violating unique key constraints:
...
-    foobar: true
-  } IN users OPTIONS { ignoreErrors: true }
+  UPDATE CONCAT('test', i)
+  WITH { foobar: true } IN users
+  OPTIONS { ignoreErrors: true }
 ```
 
-An update operation will only update the attributes specified in `document` and
+An update operation only update the attributes specified in `document` and
 leave other attributes untouched. Internal attributes (such as `_id`, `_key`, `_rev`,
```
romanatarango, line 228:
```suggestion
An update operation only updates the attributes specified in `document` and
leaves other attributes untouched. Internal attributes (such as `_id`, `_key`, `_rev`,
```

3.10/aql/operations-update.md
```diff
@@ -167,64 +219,62 @@ update non-existing documents or violating unique key constraints:
...
-Updating a document will modify the document's revision number with a server-generated value.
+`_from` and `_to`) cannot be modified and are ignored when specified in `document`.
+Updating a document modifies the document's revision number with a server-generated value.
 
 ### `keepNull`
 
-When updating an attribute with a null value, ArangoDB will not remove the attribute 
-from the document but store a null value for it. To get rid of attributes in an update
-operation, set them to null and provide the `keepNull` option:
+When updating an attribute to a `null` value, ArangoDB does not remove the attribute
```
romanatarango, line 234:
```suggestion
When updating an attribute to the `null` value, ArangoDB does not remove the attribute 
```

3.10/aql/operations-update.md
```diff
@@ -167,64 +219,62 @@ update non-existing documents or violating unique key constraints:
...
+Updating a document modifies the document's revision number with a server-generated value.
 
 ### `keepNull`
 
-When updating an attribute with a null value, ArangoDB will not remove the attribute 
-from the document but store a null value for it. To get rid of attributes in an update
-operation, set them to null and provide the `keepNull` option:
+When updating an attribute to a `null` value, ArangoDB does not remove the attribute 
+from the document but stores this `null` value. To remove attributes in an update
+operation, set them to `null` and set the `keepNull` option to `false`:
```
romanatarango, line 237: It is not clear from the text whether in this case also attributes that have been set to `null` _previously_ are deleted.

3.10/aql/operations-update.md
```diff
@@ -167,64 +219,62 @@ update non-existing documents or violating unique key constraints:
...
-  UPDATE u WITH {
-    foobar: true,
-    notNeeded: null
-  } IN users OPTIONS { keepNull: false }
+  UPDATE u WITH { foobar: true, notNeeded: null } IN users
+  OPTIONS { keepNull: false }
 ```
 
-The above query will remove the `notNeeded` attribute from the documents and update
+The above query removes the `notNeeded` attribute from the documents and update
```
romanatarango, line 244:
```suggestion
The above query removes the `notNeeded` attribute from the documents and updates
```

3.10/aql/operations-update.md
```diff
@@ -167,64 +219,62 @@ update non-existing documents or violating unique key constraints:
...
 the `foobar` attribute normally.
 
 ### `mergeObjects`
 
-The option `mergeObjects` controls whether object contents will be
-merged if an object attribute is present in both the `UPDATE` query and in the 
+The option `mergeObjects` controls whether object contents are
+merged if an object attribute is present in both, the `UPDATE` query and in the
```
romanatarango, line 250:
```suggestion
merged if an object attribute is present in both the `UPDATE` query and in the 
```

romanatarango: no comma

3.10/aql/operations-update.md
```diff
@@ -282,7 +331,7 @@ are allowed, too). These statements can refer to the pseudo-values `OLD` and `NE
 The `OLD` pseudo-value refers to the document revisions before the update, and `NEW` 
 refers to document revisions after the update.
 
-Both `OLD` and `NEW` will contain all document attributes, even those not specified 
+Both ,`OLD` and `NEW`, contain all document attributes, even those not specified
```
romanatarango, line 334:
```suggestion
Both `OLD` and `NEW` contain all document attributes, even those not specified 
```

3.11/aql/operations-replace.md
```diff
@@ -25,104 +25,171 @@ REPLACE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></co
...
+The following query replaces the document identified by the key `my_key` in the
+`users` collection, only setting a `name` attribute. The key is passed via
+the `_key` attribute alongside other attributes:
+
+```aql
+REPLACE { _key: "my_key", name: "Jon", status: "active" } IN users
+```
+
+The following query is invalid because the object does not contain a `_key`
+attribute and thus it is not possible to determine the documents that needs to
```
romanatarango, line 47:
```suggestion
attribute and thus it is not possible to determine the documents to
```

3.10/aql/operations-update.md
```diff
@@ -25,32 +25,72 @@ UPDATE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></cod
...
+The following query adds or updates the `name` attribute of the document
+identified by the key `my_key` in the `users` collection. The key is passed via
+the `_key` attribute alongside other attributes:
+
+```aql
+UPDATE { _key: "my_key", name: "Jon" } IN users
+```
+
+The following query is invalid because the object does not contain a `_key`
+attribute and thus it is not possible to determine the documents that needs to
```
romanatarango, line 46:
```suggestion
attribute and thus it is not possible to determine the documents to
```

3.11/aql/operations-update.md
```diff
@@ -25,32 +25,72 @@ UPDATE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></cod
...
+The following query adds or updates the `name` attribute of the document
+identified by the key `my_key` in the `users` collection. The key is passed via
+the `_key` attribute alongside other attributes:
+
+```aql
+UPDATE { _key: "my_key", name: "Jon" } IN users
+```
+
+The following query is invalid because the object does not contain a `_key`
+attribute and thus it is not possible to determine the documents that needs to
```
romanatarango, line 46:
```suggestion
attribute and thus it is not possible to determine the documents to
```

romanatarango (changes requested):
LGTM up to several typos

3.10/aql/operations-update.md
```diff
@@ -167,64 +219,62 @@ update non-existing documents or violating unique key constraints:
...
+Updating a document modifies the document's revision number with a server-generated value.
 
 ### `keepNull`
 
-When updating an attribute with a null value, ArangoDB will not remove the attribute 
-from the document but store a null value for it. To get rid of attributes in an update
-operation, set them to null and provide the `keepNull` option:
+When updating an attribute to a `null` value, ArangoDB does not remove the attribute 
+from the document but stores this `null` value. To remove attributes in an update
+operation, set them to `null` and set the `keepNull` option to `false`:
```
Simran-B, line 236:
```suggestion
operation, set them to `null` and set the `keepNull` option to `false`. This removes
the attributes you specify but not any previously stored attributes with the `null` value:
```

3.10/aql/operations-update.md
Simran-B: Added "This removes the attributes you specify but not any previously stored attributes with the `null` value"

3.10/aql/operations-replace.md
```diff
@@ -25,104 +25,171 @@ REPLACE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></co
 Both variants can optionally end with an `OPTIONS { … }` clause.
 
 `collection` must contain the name of the collection in which the documents should
-be replaced. `document` is the replacement document. When using the first syntax, `document` 
-must also contain the `_key` attribute to identify the document to be replaced. 
+be replaced. `document` must be an object that contains the attributes and values
```
romanatarango, line 28: Only in the second variant. In the first one, `document` just provides the `_key` attribute (as on the second variant).

BWT, I'd vote to stress (both here and in the description of `UPDATE`) that the `document` attribute of the first variant is, in fact, just another way to describe `_key`. Otherwise the use of `document` is rather counter-intuitive and even misleading: it has nothing to do with the replacing `document` (except for the value of `_key` which is implicit in the `REPLACE`/`UPDATE`call).

3.10/aql/operations-replace.md
```diff
@@ -25,104 +25,171 @@ REPLACE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></co
 Both variants can optionally end with an `OPTIONS { … }` clause.
 
 `collection` must contain the name of the collection in which the documents should
-be replaced. `document` is the replacement document. When using the first syntax, `document` 
-must also contain the `_key` attribute to identify the document to be replaced. 
+be replaced. `document` must be an object that contains the attributes and values 
+to set. All existing attributes in the stored document are removed from it and
+only the provided attributes are set (excluding system attributes).
```
romanatarango, line 30: This is the place where `UPDATE` and `REPLACE` differ and should be better stressed, e.g.:
```suggestion
to set. All existing attributes in the stored document are removed from it and
only the provided attributes are set (excluding system attributes). This distinguishes the `REPLACE` operation
from `UPDATE` where the existing attributes are not changed if not mentioned in the `UPDATE` call.
```

3.10/aql/operations-replace.md
```diff
@@ -25,104 +25,171 @@ REPLACE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></co
...
+
+### `REPLACE <document> IN <collection>`
+
+Using the first syntax, the `document` object must have a `_key` attribute with
+the document key. The existing document with this key is replaced with the
+attributes provided by the `document` object.
```
romanatarango, line 36: Again, except for the system properties. (What if the user reads only this paragraph?)

3.10/aql/operations-replace.md
```diff
@@ -25,104 +25,171 @@ REPLACE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></co
...
+
+The following query replaces the document identified by the key `my_key` in the
+`users` collection, only setting a `name` attribute. The key is passed via
```
romanatarango, line 39: and `status`

3.10/aql/operations-replace.md
```diff
@@ -25,104 +25,171 @@ REPLACE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></co
...
+### `REPLACE <keyExpression> WITH <document> IN <collection>`
 
-The following queries are equivalent:
+Using the second syntax, the document to replace is defined by the
+`keyExpression`. It can either be a string with the document key, an object
+which contains a `_key` attribute with the document key, or an expression that
+evaluates to either of these two.
+
+The following query replaces the document identified by the key `my_key` in the
+`users` collection, only setting a `name` attribute. The key is passed as
```
romanatarango, line 79: and `status`

3.10/aql/operations-replace.md
```diff
@@ -25,104 +25,171 @@ REPLACE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></co
...
 ```
 
-A replace will fully replace an existing document, but it will not modify the values
-of internal attributes (such as `_id`, `_key`, `_from` and `_to`). Replacing a document
-will modify a document's revision number with a server-generated value.
+Dynamic key expressions
+-----------------------
 
-A replace operation may update arbitrary documents:
+An `REPLACE` operation may replace arbitrary documents, using either of the two
```
romanatarango, line 116:
```suggestion
A `REPLACE` operation may replace arbitrary documents, using either of the two
```

3.10/aql/operations-replace.md
```diff
@@ -25,104 +25,171 @@ REPLACE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></co
...
-  REPLACE CONCAT('test', i) WITH { foobar: true } IN users
+  REPLACE CONCAT('test', i) WITH { name: "Paula", status: "active" } IN users
 ```
 
-The documents it modifies can be in a different collection than
-the ones produced by a preceding `FOR` operation:
+Target a different collection
+-----------------------------
+
+The documents an `REPLACE` operation modifies can be in a different collection
```
romanatarango, line 132:
```suggestion
The documents a `REPLACE` operation modifies can be in a different collection
```

3.10/aql/operations-replace.md
```diff
@@ -25,104 +25,171 @@ REPLACE <em>keyExpression</em> WITH <em>document</em> IN <em>collection</em></co
...
 FOR i IN 1..1000
-  REPLACE CONCAT('test', i) WITH { foobar: true } IN users
+  REPLACE CONCAT('test', i) WITH { name: "Paula", status: "active" } IN users
 ```
 
-The documents it modifies can be in a different collection than
-the ones produced by a preceding `FOR` operation:
+Target a different collection
+-----------------------------
+
```
romanatarango, line 131:
```suggestion
Recall the example:
```aql
FOR u IN users
  REPLACE u WITH { name: CONCAT(u.firstName, " ", u.lastName), status: u.status } IN users
```
Although the `u` variable holds a whole document, it is used only to define the `_key` attribute in the `REPLACE` operation. The latter then finds a document defined by the `_key` attribute in the collection defined in the `IN` clause and replaces it with the contents of the `REPLACE` clause. Of course, if `FOR` iterates over the same collection, the document held in the variable `u` will be found and replaced. However, `FOR` can also iterate over another collection:
```aql
FOR u in users
   REPLACE u WITH { name: CONCAT(u.firstName, " ", u.lastName), status: u.status } IN backup
```
Then the replaced document is different from `u` (it is in another collection) and only shares the `_key` attribute with it.
```

**Review by romanatarango — CHANGES_REQUESTED** [review]
I think the fine distinction between the documents used to define a `_key` and those to be updated/replaced should be explained in more details.

File: 3.10/aql/operations-replace.md (hunk 13 above)
Comment (Simran-B) [reply 3.10/aql/operations-replace.md:28]: > Only in the second variant

I don't follow. With both syntaxes, `document` must be an object. This is not optional. If you use the first syntax, then this object must have a `_key` attribute. Other than that, the object can be empty, but you do commonly specify attributes - or you end up with an empty document (except system attributes).

Comment (romanatarango) [reply 3.10/aql/operations-replace.md:28]: > > Only in the second variant
> 
> I don't follow. With both syntaxes, `document` must be an object. This is not optional. If you use the first syntax, then this object must have a `_key` attribute. Other than that, the object can be empty, but you do commonly specify attributes - or you end up with an empty document (except system attributes).

It was my mistake, sorry for the confusion.

File: 3.10/aql/operations-replace.md [hunk 21]
```diff
@@ -1,18 +1,26 @@
 ---
 layout: default
-description: The REPLACE keyword can be used to completely replace documents in a collection
+description: >-
+  You can use `REPLACE` operations to completely replace documents with all
+  their attributes in a collection
```
romanatarango, line 5:
```suggestion
  their non-system attributes in a collection
```

Simran-B: This is getting out of hand... The part about system attributes was merely a side note, and it's mentioned five times on the page now. I don't see a need to mention it in the lead paragraph, too, which is only meant as a synopsis that can be inaccurate for the sake of providing a general overview. It's also incorrect to say that non-system attributes are excluded (_from and _to are system attributes but can be modified for a long time).

romanatarango (approved):
LGTM

nerpaula (approved):

3.10/aql/operations-replace.md
romanatarango: I see no harm in repeating the same information to avoid that the user misunderstands it. This is not a poem, it doesn't need to be nice, it has to be unambiguous. Not every user reads the whole page every time she needs to recall something, I guess most people never read the whole page. I want to avoid that a user reads the synopsis ("replace documents with all their attributes"), tries to replace all the attributes (how many people would check further?), this doesn't work, the user searches for a bug in her code for hours, asks the support - just to learn that the sentence in the synopsis is inaccurate.

My suggestion does not say that system (you mean system, not non-system) attributes are excluded, it says that non-system attributes are replaced. It doesn't say anything about system attributes at all (a simplification for the synopsis but nothing incorrect).

Simran-B: I meant to write:
> It's incorrect to say that system attributes are excluded

"It replaces documents with all their non-system attributes" or "It replaces documents with all their attributes excluding system attributes" is the same statement. Depending on the definition of system attributes,
- every attribute that starts with an underscore is a system attribute
- only top-level attributes that start with an underscore are system attributes
- only specific attributes are system attributes, namely `_key`, `_id`, `_rev`, `_from`, and `_to`, for which there is a special treatment in code and the API
- only specific attributes are system attributes, namely `_key`, `_id`, and `_rev`, which are immutable (`_key`), virtual attributes (`_id`), or not user-controlled (`_rev`).

This is not properly defined, and you can find inconsistent behavior concerning system attributes in ArangoDB. If you don't expect users to read more than the subheading, then we need to clearly state what it does and doesn't replace in the synopsis. It does replace _from, it does replace _to (if provided by the user), it does not replace the other three, but that is not directly related to the AQL UPDATE and REPLACE operations.

PR #1177, opened 2022-11-09
DOC-229 | In-source detailed explanations of startup options

Simran-B:
Upstreams:
- https://github.com/arangodb/arangodb/pull/17473
- https://github.com/arangodb/enterprise/pull/1136
- https://github.com/arangodb/arangodb/pull/17549
- https://github.com/arangodb/enterprise/pull/1141

The commits "Temp: Add generated startup option files" are to be reverted/removed from the branch history, and the normal example generation should be used to generate the startup option JSON files.

TODO:
- Change links from `programs-arangod-arangosearch.html` to `programs-arangod-options.html#arangosearchcolumns-cache-limit` once the other PRs are merged

PR #1185, opened 2022-11-11
Adjust docs about spillover feature

Simran-B:
Correct a statement and unify with startup option description.

To be merged after #1177 (check for conflicts / hidden broken links due to redirects).
~Change `programs-arangod-query.html#aql-query-with-spilling-input-data-to-disk`
to `programs-arangod-options.html#--tempintermediate-results-path`.~


cpjulia (approved):
LGTM

nerpaula (approved):

PR #1190, opened 2022-11-16
Partially revert #1175 because `primarySortCache` will be released in v3.9.6

Dronplane:
~~Upstream: https://github.com/arangodb/arangodb/pull/17602 (adds missing `cache` option to storedValues)~~

Too late to include upstream PRs for v3.9.5, added the changes to #1191 for v3.9.6 instead.

PR #1191, opened 2022-11-16
DOC-266 | ArangoSearch column cache (3.10, 3.11), restore `primarySortCache` & add `primaryKeyCache` (3.9)

Dronplane:
See #1190 

Upstreams:
- 3.9: https://github.com/arangodb/arangodb/pull/17603
- 3.9 (minor follow-up) https://github.com/arangodb/arangodb/pull/17685
- 3.10: https://github.com/arangodb/arangodb/pull/17684
- 3.11: https://github.com/arangodb/arangodb/pull/17683

Feature PRs:
- 3.9: https://github.com/arangodb/arangodb/pull/17413 / https://github.com/arangodb/arangodb/pull/17599
- 3.10: https://github.com/arangodb/arangodb/pull/17668
- devel https://github.com/arangodb/arangodb/pull/17628

~Documentation changes to be forward-ported from 3.9 to 3.10 and 3.11.~

~Change `programs-arangod-arangosearch.html` to `programs-arangod-options.html#--arangosearchcolumns-cache-limit` once #1177 is merged.~


Simran-B:
Are `primarySortCache` and `primaryKeyCache` mutable?

3.9/arangosearch-performance.md
```diff
@@ -339,3 +339,28 @@ used. You can create custom Analyzers without this feature to disable the
...
 still matches your expectations without normalization. It is recommended to
 use normalization for a good scoring behavior.
+
+## Primary key caching
+
+<small>Introduced in: v3.9.6</small>
+
+You can set the `primaryKeyCache` View property to `true` to always cache the
+the primary key columns in memory. This can improve the performance of queries
+that return many documents:
```
Simran-B, line 349: It would be helpful to better explain the use cases for this option, when it can help in particular, with a small example.

Dronplane:
> Are `primarySortCache` and `primaryKeyCache` mutable?

No

3.9/arangosearch-performance.md
Dronplane: The same story as with `"cache":true` for fields. But this time it is about document ids (LocalDocumentId) to get the actual doc for SEARCH.  The only difference that this field used only when producing results. E,g, if there is SORT + LIMIT than less documents will be actually produced than scanned. So this option has less effect on the performance in case of query with "LIMIT 10"

Simran-B: I see. Do you think this description is okay?
> This can improve the performance of queries that return many documents, making it faster to map document IDs in
the index to actual documents

Dronplane: Yes

Simran-B:
https://jenkins.arangodb.biz/view/Documentation/job/arangodb-ANY-examples/2653/

Simran-B (approved):

PR #1192, opened 2022-11-17
DOC 192 / ArangoGraph multi-region backups

nerpaula:
preview https://deploy-preview-1192--zealous-morse-14392b.netlify.app/docs/3.10/arangograph/backups.html


nikhil-varma (changes requested):

arangograph/backups.md
```diff
@@ -38,20 +46,58 @@ It's also possible to create a backup on demand. To do this, click **Back up now
 
 ![Back up Now Dialog](images/arangograph-back-up-now-dialog.png)
 
+If you want to manually copy a backup to a different region than the default
+one, first ensure that the **Upload backup to storage** option is enabled.
+Then, highlight the backup row and use the
+**Copy backup to a different region** button from the **Actions** column. 
+Once the copy of the default backup is created in the region of your choice,
+you can copy it to a different region as well. The source backup ID from
```
nikhil-varma, line 54: Ideally, the default backup allows creating multiple copies of itself into multiple regions one at a time if done manually. Maybe this point `Once the copy of the default backup is created in the region of your choice,
you can copy it to a different region as well` can be misleading? What do you think?

nikhil-varma (changes requested):

arangograph/backups.md
```diff
@@ -14,6 +14,7 @@ To backup data in ArangoGraph for an ArangoDB installation, navigate to the
 
 There are two ways to create backups. Create periodic backups using a
 **Backup policy**, or create a backup manually.
+Both ways allow you to create backups in multiple regions as well.
```
ansoboleva, line 17: Please say that this feature is available for enterprise tier only.

arangograph/backups.md
```diff
@@ -23,12 +24,19 @@ due, observe the schedule section.
...
 (or any policy), highlight it in the row above and hit the pencil icon.
 
 ![Edit Backup Policy](images/arangograph-edit-backup-policy.png)
 
-These backups are not automatically uploaded.
+These backups are not automatically uploaded. To enable this, use the
+**Upload backup to storage** option and choose a retention period that
+specifies how long backups are retained after creation. 
+
+When the **Upload backup to storage** option is enabled for a backup policy,
```
ansoboleva, line 36: Please say that this feature is available for enterprise tier only.

arangograph/backups.md
```diff
@@ -38,20 +46,58 @@ It's also possible to create a backup on demand. To do this, click **Back up now
...
+**Upload backup to cloud storage** option is enabled.
+
+![Multiple Region Backup](images/arangograph-multi-region-backup.png)
+
 ## How to restore backups
 
 To restore a database from a backup, highlight the desired backup and click the restore icon.
 
 {% hint 'warning' %}
 All current data will be lost when restoring.
```
ansoboleva, line 99: What is meant by this exactly? (All current data will be lost). If some data loss may happen, we also need to say this in the UI warning - this currently only says the database will be unavailable for some time. Please check and adjust the UI message/docs as needed.

ansoboleva (changes requested):
please see my comments

arangograph/backups.md
nerpaula: It is already mentioned here https://deploy-preview-1192--zealous-morse-14392b.netlify.app/docs/3.10/arangograph/backups.html#multi-region-backups

nerpaula: but we can add it here and below as well to avoid confusion

arangograph/backups.md
```diff
@@ -38,20 +46,58 @@ It's also possible to create a backup on demand. To do this, click **Back up now
 
 ![Back up Now Dialog](images/arangograph-back-up-now-dialog.png)
 
+If you want to manually copy a backup to a different region than the default
+one, first ensure that the **Upload backup to storage** option is enabled.
+Then, highlight the backup row and use the
+**Copy backup to a different region** button from the **Actions** column.
```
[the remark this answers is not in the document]
nerpaula: enterprise only

arangograph/backups.md
nerpaula: @nikhil-varma maybe you can answer this question?

Robert-Stam: Assume you create a backup at t1, then at t2 (after already inserting new data) you are restore the backup from t1, the data inserted after that time will be 'lost'. if you want to go back to that data you can create a backup just before you restore a backup, that way, you can go back to t2 again.

arangograph/backups.md
```diff
@@ -66,17 +112,23 @@ During restore the deployment is temporarily not available.
 
 {% hint 'info' %}
 The cloned deployment will have the exact same features as the previous
-deployment including node size, model, cloud provider & region. The data
-contained in the backup will be restored to this new deployment.
+deployment including node size, model, and cloud provider. The region
+can stay the same or you can select a different one. 
+The data contained in the backup will be restored to this new deployment.
```
Robert-Stam, line 117: For restoring as quick as possible, we recommend to create a deployment in the same region as where the backup resides, to avoid data transfer cross regions.

Robert-Stam (commented):
LGTM

arangograph/backups.md
ansoboleva: or make it a link then to the section where it's mentioned

arangograph/backups.md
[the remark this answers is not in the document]
ansoboleva: Then we need to say it. Not necessarily make a note, but at least mention in parentheses that it's enterprise only. Otherwise we're forcing users to create a connection between seemingly unconnected text sections. Users can only read one of the sections and not the entire page, but they still should be able to get the necessary info.

ansoboleva (approved):

PR #1193, opened 2022-11-18
DOC-12 | Improve PRUNE docs / UDFs and DB-Server only functions not allowed in PRUNE

jsteemann:
- [x] AQL UDFs no more allowed in PRUNE expressions and the functions must be runnable on DB-Servers (also applies to `FILTER` conditions that can be pulled into the traversal on a DB-Server)
- [x] Add a list of the built-in Coordinator-only functions? Wording/list can be copied from [computed values](https://www.arangodb.com/docs/stable/data-modeling-documents-computed-values.html#computed-value-expressions). We should consider annotating each AQL function in the reference with this information, also see DOC-9
- [x] Improve [PRUNE docs](https://www.arangodb.com/docs/stable/aql/graphs-traversals.html#pruning) with better/more examples (repeating PRUNE conditions in FILTER, possibly with negation) and point out pitfalls (`null` edge, prunes at all levels including start vertex, different/inverted logic compared to FILTER, the current node/edge/path is included in the result where the condition is met)


3.8/aql/extending.md
```diff
@@ -43,6 +43,8 @@ Rule of thumb is, the closer the UDF is to your final `RETURN` statement
 
 When used in clusters, UDFs are always executed on the
 [Coordinator](../architecture-deployment-modes-cluster-architecture.html).
+Queries that would push UDF execution to DB-Servers, e.g. if a UDF is used 
+inside a traversal PRUNE condition, will abort with a parse error.
```
jsteemann, line 47:
```suggestion
Queries that would push UDF execution to DB-Servers will abort with a parse error.
UDFs cannot be used in traversal PRUNE conditions at all.
```

3.9/aql/extending.md
```diff
@@ -43,6 +43,8 @@ Rule of thumb is, the closer the UDF is to your final `RETURN` statement
 
 When used in clusters, UDFs are always executed on the
 [Coordinator](../architecture-deployment-modes-cluster-architecture.html).
+Queries that would push UDF execution to DB-Servers, e.g. if a UDF is used 
+inside a traversal PRUNE condition, will abort with a parse error.
```
jsteemann, line 47:
```suggestion
Queries that would push UDF execution to DB-Servers will abort with a parse error.
UDFs cannot be used in traversal PRUNE conditions at all.
```

Simran-B:
Functions like `COLLECTIONS()` that cannot run on DB-Servers are generally disallowed, including single server and not only cluster deployments. This is good for deployment mode parity.

What I will add to this PR:
- A description of the limitations to the PRUNE docs
- Remarks about PRUNE, that it applies at all depths including the starting vertex and also the depths between 1 and the min depth of the traversal
- Common pitfalls, especially the first edge which is null
- Examples of PRUNE + FILTER, using a prune variable, with and without negation for different usage scenarios

Simran-B (approved):

PR #1202, opened 2022-11-24
Make two deprecated `--server.disable-authentication*` server startup options obsolete

Simran-B:
Upstream: https://github.com/arangodb/arangodb/pull/17636


nerpaula (approved):
LGTM

PR #1205, opened 2022-11-29
Update overview diagram for ArangoGraph, add source SVG

Simran-B:
The image URLs need to be purged from the Cloudflare cache after the deployment to Pantheon, or users will keep seing the older version


nerpaula (approved):
LGTM

PR #1214, opened 2022-12-06
DOC-278 | Edge cache refilling (experimental) [3.9]

Simran-B:
Feature PR: https://github.com/arangodb/arangodb/pull/17705

Upstream (DocuBlocks, Metrics, startup options): https://github.com/arangodb/arangodb/pull/17758

(The 3.9.6 release process has already started so we may need to merge this docs PR and leave the upstream PR open for later)

To be forward-ported to 3.10(.2) with minimal changes (introduced in versions). Forward port to 3.11 needs adjustments because the feature also supports persistent indexes with in-memory hash cache and not only the edge cache.

### DIFF
--- a/3.9/aql/operations-insert.md
+++ b/3.9/aql/operations-insert.md
@@ -168,6 +168,16 @@ FOR doc IN collection
   OPTIONS { exclusive: true }
 ```
 
+### `refillIndexCaches`
+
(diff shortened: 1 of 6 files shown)


jsteemann (approved):
LGTM

nerpaula (approved):
LGTM

PR #1218, opened 2022-12-07
DOC-67 | Only built-in JS AQL functions use _aql namespace

Simran-B:
V8() is the only remaining function with a JavaScript implementation, living in the default namespace. Native functions don't have a namespace. User-defined functions require one or more namespaces.


jsteemann (approved):
LGTM

ansoboleva (approved):

PR #1220, opened 2022-12-13
Explain AF cluster differences.

joerg84:
Added more details on the differences between AF and cluster architecture.


joerg84:
@nerpaula we would need this also in all other versions.

nerpaula:
> @nerpaula we would need this also in all other versions.

@joerg84 added this to the other versions as well.

PR #1200, opened 2022-11-23
DOC-227 | Remove tutorials

Simran-B:
- [x] Delete the tutorials link collection
- [x] Delete K8s tutorials, but add a minikube example and some other useful information to "Using the ArangoDB Kubernetes Operator"
- [ ] Delete the DC2DC tutorial, integrate useful information into other pages
- [x] #1180
- [ ] Delete the ArangoDB Starter tutorial, integrate useful information into other pages

PR #1224, opened 2022-12-27
kube-arangodb: Revert ArangoLocalStorage apiVersion from v1 to v1alpha

Simran-B:
This was incorrectly changed in #1200


nerpaula (approved):
LGTM

PR #1225, opened 2022-12-27
DOC-30 | DOC-33 | Clarifications around ArangoSearch

Simran-B:
Open points:
1. (Re)moving stop word files no longer appears to prevent the server from starting, but now you get inconsistencies. If the files are missing, all tokens get indexed. If they are there, some tokens may get discarded. It depends on the availability at the moment of the processing/indexing.
2. The German Snowball stemming algorithm does some weird things. Out of scope for our docs?
3. Does the `text` Analyzer tokenize first, then apply normalizations, and finally stemming?
4. Concrete example for exponential(?) complexity of certain queries? https://github.com/arangodb/docs/issues/296
5. Is `storedValues` needed in addition to `fields` because the index entries of `fields` do not match the actual document attributes, i.e. they cannot cover projections?
6. Analyzer input and output types, esp. in a pipeline (DOC-33)
   - "identity" in a pipeline isn't useful but it's allowed (the web interface hides this option)
   - Why is "analyzer" with a non-string return value incompatible with other Analyzers that are text-based? They do accept booleans and numbers in other contexts (but do not process them).
7. The `delimiter` Analyzer seems to support quoting delimited fragments, allowing you to escape the delimiter, e.g. `foo;"bar;baz";moo` is split into `foo`, `bar;baz`, and `moo` with `,` as delimiter. Is this a proper feature?


MBkkt (commented):
View part looks ok for me.

About analyzers I just don't know

3.10/analyzers.md
```diff
@@ -141,11 +158,39 @@ Analyzer  /  Capability                   | Tokenization | Stemming | Normalizat
...
+The `norm`, `text`, and `segmentation` Analyzer types allow you to convert the
+input text to all lowercase or all uppercase for normalization purposes, namely
+case insensitive search. Case folding is not supported. Also see
+[Case Mapping](https://unicode-org.github.io/icu/userguide/transforms/casemappings.html){:target="_blank"}
+in the ICU documentation.
+
+The `norm` and `text` Analyzer types also allow you to convert characters with
+diacritical marks to the base characters. This normalization enables
+accent-insensitive search.
+
```
Dronplane, line 193: Are there plans to describe Stemming and n-grams  here?

3.10/analyzers.md
```diff
@@ -801,9 +869,14 @@ attributes:
...
-will make the creation fail. These Analyzers require additional postprocessing
-and can only be applied to document fields directly.
+- You cannot use Analyzers of the types `geopoint` and `geojson` in pipelines.
+  These Analyzers require additional postprocessing and can only be applied to
+  document fields directly.
+- The output data type of an Analyzer needs to be compatible with the input
+  data type of the next Analyzer in the chain. For example, the `aql` Analyzer
+  has a `returnType` property, and if you set it to `number` or `bool`, the
+  subsequent Analyzer in the pipeline needs to support this data type as input.
+  Most Analyzers expect string inputs and are thus incompatible.
```
Dronplane, line 879:
```suggestion
  Most Analyzers expect string inputs and are thus incompatible with such setup.
```
Otherwise that reads like "Just all anaylzers are incompatible". That looks a bit strange. I'd say more - why mention this setting of "aql" analyzer here? It should be described in the corresponding analyzer section with note that setting custom return type will prevent this analyzer from being used in the middle of pipeline.

3.10/analyzers.md
```diff
@@ -849,8 +922,8 @@ An Analyzer capable of removing specified tokens from the input.
 
 It uses binary comparison to determine if an input token should be discarded.
 It checks for exact matches. If the input contains only a substring that
-matches one of the defined stopwords, then it is not discarded. Longer inputs
-such as prefixes of stopwords are also not discarded.
+matches one of the defined stop words, then it is not discarded. Longer inputs
```
Dronplane, line 925: I'd say "stopwords" should be used as it matches what we have in the options. No strong opinion though.

3.10/aql/operations-search.md
```diff
@@ -6,10 +6,14 @@ title: The SEARCH operation in AQL
 SEARCH
 ======
 
-The `SEARCH` keyword starts the language construct to filter Views.
+The `SEARCH` keyword starts the language construct to filter Views, accelerated
+by the underlying indexes. It guarantees to use these indexes for an efficient
+execution plan. If you use the `FILTER` keyword for Views, no indexes are
```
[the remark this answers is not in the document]
Simran-B: You can post-filter everything, but it obviously performs badly if you fetch everything from the View and then filter afterwards.

3.10/analyzers.md
Simran-B: Out-of-scope for this PR but I added it to the list of https://arangodb.atlassian.net/browse/DOC-242

3.10/analyzers.md
Simran-B: The correct spelling appears to be "stop words", which is what we should use if you refer to the concept, and `stopwords` if it's a code value because that's how the API is.

3.10/analyzers.md
Simran-B, line 879:
```suggestion
  data type of the next Analyzer in the chain. The `aql` Analyzer, in particular,
  has a `returnType` property, and if you set it to `number` or `bool`, the
  subsequent Analyzer in the pipeline needs to support this data type as input.
  Most Analyzers expect string inputs and are thus incompatible with such a setup.
```

Dronplane:
Some open points:
_1. (Re)moving stop word files no longer appears to prevent the server from starting, but now you get inconsistencies._ 
Could you please elaborate a bit more. As I see in the code just missing  stopwords in the  "default path" e.g. CWD is ignored. But failure to load them from explicitly specified location is still an error. 

_2.The German Snowball stemming algorithm does some weird things. Out of scope for our docs?_
I guess that is out of scope for our docs. We could not describe all the pitfalls for all languages. I guess we already mention in our docs that we rely on snowball stemming.  IMO that is enough.

_3. Does the text Analyzer tokenize first, then apply normalizations, and finally stemming?_
Accent removals -> case converstion->stopwords filtering->stemming

_4. Concrete example for exponential(?) complexity of certain queries?_
Suggets to  just note that if profiling shows much time spent in the "optimization" stage  and query has more than 5 conjunctions of disjunctions - it is a good reason to try disabling DNF transformation.

_5. Is storedValues needed in addition to fields because the index entries of fields do not match the actual document attributes, i.e. they cannot cover projections?_
Yes, for inverted index toredValues are used to serve projections


_6. Analyzer input and output types, esp. in a pipeline (DOC-33)
"identity" in a pipeline isn't useful but it's allowed (the web interface hides this option)
Why is "analyzer" with a non-string return value incompatible with other Analyzers that are text-based? They do accept booleans and numbers in other contexts (but do not process them)._

For identity - nothing prevents using it in the pipeline. So why bother with explicitly forbiding? 
For text-based analyzers accepting non-strings - you mean TOKENS function? Or indexing? They don't accept such types. They are just silently substituted by corresponding primitive analyzers under the hood. For pipeline we don't have this substitution algorithm. We could think about adding such features in future if there would be real need.


_7. The delimiter Analyzer seems to support quoting delimited fragments, allowing you to escape the delimiter, e.g. foo;"bar;baz";moo is split into foo, bar;baz, and moo with , as delimiter. Is this a proper feature?_

Yes, that is the feature

Simran-B:
2. We do link to the Snowball website, so this is done.
3. Alright, I swapped accent removal and case conversion steps in the PR.
4. We already have a page about ArangoSearch performance, which talks about condition optimization and the option to disable it. This might be sufficient.
5. ✔️ 
6. Yeah, my question was basically aiming at that substitution and it being missing for the pipeline Analyzer. I can't think of a good use case from the top of my head.
7. Added something about escaping of delimiters

nerpaula (approved):

PR #1229, opened 2023-01-04
ArangoGraph rebranding - replace remaining images

nerpaula:
All images are now updated.


Simran-B:
We can't replace `oasis-org-invite-email-verify.png` yet because the email still contains "Oasis" and the old logo

Robert-Stam (approved):
LGTM

PR #1231, opened 2023-01-04
DOC-52 | DOC-55 | Clarifications around replication factor and write concern

Simran-B:
- `replicationFactor` cannot be higher than DB-Server count, except if you set `enforceReplicationFactor` to `false` on collection creation (not possible later)
- `replicationFactor` is clamped by `--cluster.min-replication-factor` and `--cluster.max-replication-factor`
- `replicationFactor` defaults to `--cluster.default-replication-factor` for regular collections and to `--cluster.system-replication-factor` for system collections
- `writeConcern` sets the number of minimum in-sync shard replicas (i.e. the total number of copies, not the minimum follower count)
- Add more functions that prevent the OneShard optimizations (`canRunOnDBServerOneShard: false`)
- Some cleanup


3.10/administration-cluster.md
```diff
@@ -20,45 +20,59 @@ Enabling synchronous replication
 For an introduction about _Synchronous Replication_ in Cluster, please refer
 to the [_Cluster Architecture_](architecture-deployment-modes-cluster-architecture.html#synchronous-replication) section. 
 
-Synchronous replication can be enabled per _collection_. When creating a
-_collection_ you may specify the number of _replicas_ using the
-*replicationFactor* parameter. The default value is set to `1` which
-effectively *disables* synchronous replication among _DB-Servers_. 
+You can enable synchronous replication per _collection_. When you create a
+_collection_, you may specify the number of _replicas_ using the
```
neunhoef, line 24: I would add an explicit comment here, that this number includes the leader and all followers. This is a common misunderstanding and cannot be emphasized often enough. Something along the lines of:
"The number of replicas includes the leader (the master copy) as well as all the followers (redundancy copies)."
Just a suggestion, maybe you find this clear enough, then just feel free to ignore.

3.10/architecture-replication.md
```diff
@@ -28,62 +28,68 @@ Synchronous replication only works within an ArangoDB Cluster and is typically
...
-Transactions are honored in replication, i.e. transactional write operations will 
+Transactions are honored in replication, i.e. transactional write operations 
 become visible on _Followers_ atomically.
 
-As all write operations will be logged to a Leader database's _write-ahead log_, the 
-replication in ArangoDB currently cannot be used for write-scaling. The main purposes 
-of the replication in current ArangoDB are to provide read-scalability and "hot backups" 
+As all write operations are logged to a Leader database's _write-ahead log_, the 
+replication in ArangoDB cannot be used for write-scaling. The main purposes 
+of the replication are to provide read-scalability and "hot backups"
```
neunhoef, line 76: The usage of the term "hot backups" here is misleading, since we have used this for something else, since this was written. How about "hot standby servers" or something like this?

neunhoef (approved):
LGTM. See some small comments. Note that they are only in the 3.10 part, since the 3.11 part seems to be a perfect copy.

3.10/administration-cluster.md
Simran-B: Good call!

neunhoef (approved):
LGTM

nerpaula (approved):

PR #1233, opened 2023-01-05
DOC-275 | Clarify how to add additional servers to starter cluster

Simran-B:


neunhoef (approved):
LGTM. What about 3.9 and 3.8? Should this be "backported" there, too?

nerpaula (approved):

PR #1236, opened 2023-01-06
DOC-237 | Reduce duplicate content about data (modification) queries

Simran-B:
Integrate modification query example content into data queries page. The initial plan was to dissolve the data queries page too, but upon closer inspection, it contains some relevant reference documentation that is hard to integrate elsewhere - we would need to duplicate the content because we have separate pages about UPDATE and REPLACE, for example. There is also no good overview about writing queries without this page.


3.11/aql/data-queries.md
```diff
@@ -120,228 +131,430 @@ You may provide a key for the new document; if not provided, ArangoDB creates on
...
+      {
+        "attributeToAlter": "keep this value",
+        "filterByMe": false
+      }
+    ]
+  }
+]
+```
+
+To improve the query's performance, you can only updating documents if there is
```
nerpaula, line 420:
```suggestion
To improve the query's performance, you can only update documents if there is
```

PR #1239, opened 2023-01-11
DOC-305 | Streaming cursors can report warnings in respective batch

Simran-B:
Currently, this is limited to single server deployments as per a comment in https://arangodb.atlassian.net/browse/APM-570

Upstream: https://github.com/arangodb/arangodb/pull/17856


Simran-B:
There are technical issues with the implementation. It is unclear if or when they can be resolved. This PR is thus on-hold.

nerpaula (approved):

PR #1242, opened 2023-01-12
Cloud migration tool - added env variables and limitations

nerpaula:


arangograph/cloud-migration-tool.md
```diff
@@ -245,6 +245,31 @@ error code `11` (ERROR_FORBIDDEN).
...
+bigger) amount of resources (CPU, RAM) than your cluster. Otherwise, the
+migration process might get stuck or require manual intervention. This is closely
+connected to the type of data you have and how it is distributed between shards
+and collections.
+
+In general, the most important parameters are:
+- Total number of leader shards
+- The amount of data in bytes per collection
+
+Both parameters can be retrieved from the ArangoDB Web Interface.
```
nikita-vanyasin, line 271: The main idea of updating migration agent documentation was to have a specific numbers of what we tested - to provide expectations for customers on how much data he can migrate.

From [draft](https://github.com/arangodb/arangosync#restrictions-and-limitations):
```
The largest dataset we tested had 5TB of data and 3800 leader shards. Some collections were as big as 250GB of data.

We expect the migration to work fine even with bigger datasets, but if you have use-case with bigger numbers please reach out to us, so we will perform more tests with datasets similar to yours.
```

@joerg84 please let us know if this is not needed.

mattcarabine: I would not set the expectation that we'd definitely perform more tests for anyone that gets in contact with us with a larger dataset. 
Agree with the rest of your suggestion @nikita-vanyasin.

nerpaula: done.

arangograph/cloud-migration-tool.md
```diff
@@ -245,6 +245,37 @@ error code `11` (ERROR_FORBIDDEN).
 The `--source.server-mode` option allows you to specify the desired server mode.
 Allowed values are `readonly` or `default`.
 
+### Supported environment variables
+
+The `arangosync-migration` tool supports the following environment variables:
+
+- `$OASIS_API_KEY`
```
joerg84, line 252: @nikita-vanyasin should we rename these these (i.e., non Oasis)?
We can keep supporting the old ones for backwards comp, but at least in the docs I would mention the new ones...

joerg84 (approved):

arangograph/cloud-migration-tool.md
nikita-vanyasin: @joerg84 sure! I'll create a ticket for migration tool.

nikita-vanyasin (approved):

PR #1246, opened 2023-01-16
Move OneShard content under Deployment

nerpaula:


3.10/deployment-oneshard.md
```diff
@@ -0,0 +1,311 @@
...
+# OneShard
+
+{{ page.description }}
+{:class="lead"}
+
+{% include hint-ee-arangograph.md feature="The OneShard option" %}
+
+A OneShard deployment offers a practicable solution that enables significant
+performance improvements by massively reducing cluster-internal communication
+and allows running transactions with ACID guarantees on shard leaders.
```
Simran-B, line 16: This is almost identical to the lead paragraph. I would add the additional point about ACID properties to the lead paragraph and remove this paragraph here.

PR #1262, opened 2023-02-07
ArangoGraph | Single-Sign On with Okta

nerpaula:
Update ArangoGraph Single-Sign On (SSO) feature:
- separate sections for SAML 2.0 and SCIM
- step-by-step procedures on how to configure SAML 2.0 and SCIM using Okta

preview: https://deploy-preview-1262--zealous-morse-14392b.netlify.app/docs/3.10/arangograph/single-sign-on.html


arangograph/single-sign-on.md
```diff
@@ -25,17 +44,98 @@ can sync the information about user resources in real-time:
...
+3. In the **Create a new app integration** dialog, select **SAML 2.0**.
+
+   ![ArangoGraph Create Okta App Integration](images/arangograph-okta-create-integration.png)
+4. In the **General Settings**, specify a name for your integration.
+
+   ![ArangoGraph Okta Integration Name](images/arangograph-okta-integration-name.png)
+5. Click **Next**.  
+6. Configure the SAML settings:
+   - For **Single sign-on URL**, use `https://auth.arangodb.com/login/callback?connection=ORG_ID`
+   - For **Audience URI (SP Entity ID)**, use `urn:auth0:arangodb:ORG_ID`
```
mattcarabine, line 82: We should give some examples here, for example "if your organization ID is 123, here are the values you'd use", since AFAIK these values won't work as-is, we need to substitute.

arangograph/single-sign-on.md
```diff
@@ -25,17 +44,98 @@ can sync the information about user resources in real-time:
...
+   is displayed under the **Sign On** tab.
+2. To finalize the SSO configuration, you must perform the SAML setup.
+3. Click **View SAML setup instructions**.
+
+   ![ArangoGraph Okta SAML Setup](images/arangograph-okta-saml-setup.png)
+4. The setup instructions include the following items:
+   - Identity Provider Single Sign-On URL
+   - Identity Provider Issuer
+   - X.509 Certificate
+5. Copy the IdP settings, download the certificate, and share them with the
```
mattcarabine, line 116: So two things here:
1. We should be prescriptive on exactly what they should copy (is there a "copy IdP settings" button)? 
2.  We should give some advice on how they can share these with us. There are two paths as I see it, one where they're already engaged with us, and a path where they're trying to self-service. Perhaps we can just say "share them with the ArangoGraph team via an ArangoGraph Support ticket".

arangograph/single-sign-on.md
```diff
@@ -9,9 +9,28 @@ title: Single Sign-On (SSO)
...
 {% endhint %}
 
-ArangoGraph provides support to enable Single Sign-On (SSO) authentication using
-SAML 2.0 with SCIM provisioning. This enables you to propagate to ArangoGraph any
-user access changes by using the dedicated API.
+ArangoGraph provides support to enable **Single Sign-On** (SSO) authentication
+using **Security Assertion Markup language 2.0** (SAML 2.0) with the
+**System for Cross-domain Identity Management** (SCIM) provisioning. 
+This enables you to propagate to ArangoGraph any user access changes by using
+the dedicated API.
```
mattcarabine, line 16: Can we change how this is positioning SCIM?

Having read through the docs it reads as though you have to use SCIM to setup SSO, whereas the reality is that we don't want people to use SCIM if we can avoid it, it's an optional extra that most do not need.
Perhaps we can have the SSO SAML page (this one) with a child page of SCIM, or just some other way to get the point across to users that "If you want to configure SSO, do these steps then stop here. If you then want SCIM on top, this is where you can go".

arangograph/single-sign-on.md
```diff
@@ -9,9 +9,28 @@ title: Single Sign-On (SSO)
...
+
+{% hint 'info' %}
+To enable the Single Sign-On (SSO) feature, please
+[contact us](https://www.arangodb.com/contact){:target="_blank"}.
```
mattcarabine, line 20: Can we change this to pointing them towards their ArangoDB account team, or to an ArangoGraph Support ticket please?

mattcarabine (changes requested):
Left some individual comments but the main gist is that we need to change the way that this SSO stuff flows.

We need to heavily de-emphasise SCIM and more cleanly separate SCIM and SAML. It's fine that all of the SAML configuration steps are a pre-requisite to enable SCIM, but right now it looks like they're fully combined. I'd suggest something like a completely different page for SCIM so that we don't mix them.

e.g. a page that goes through all the steps of configuring SAML SSO, then a "If you're interested in SCIM, you can also look here" as a follow-on.

As it reads today, everybody will be enabling SCIM and we do not want that.

arangograph/single-sign-on.md
```diff
@@ -25,17 +44,98 @@ can sync the information about user resources in real-time:
...
+
+## Configure SAML 2.0 and SCIM using Okta
+
+You can enable SSO for your ArangoGraph organization using Okta as an Identity
+Provider (IdP). For more information about Okta, please refer to the
+[Okta Documentation](https://help.okta.com/en-us/Content/index.htm?cshid=csh-index){:target="_blank"}.
+
+### Create the SAML app integration in Okta
+
+1. Go to Okta and select **Applications** from the left sidebar menu.
```
Simran-B, line 71: Can we make Okta a link to their dashboard or whatever they have to make clearer where to go?

arangograph/single-sign-on.md
Simran-B: And it would be good to instruct users not to send anything over the wire unencrypted, e.g. via normal email.

arangograph/single-sign-on.md
nerpaula: I don't think Okta has generic dashboard links, but we can add that they need to sign into Okta first

arangograph/single-sign-on.md
nerpaula: 1. There is no copy button for IdP settings (you need to manually copy both fields). However, there is a "Download X.509 certificate" button. 
2. We can instruct them to always share these via an ArangoGraph Support ticket.

arangograph/single-sign-on.md
nerpaula: Good to know that we want to avoid enabling SCIM. In this case, we can move the sections "About SCIM" and "How to enable SCIM in Okta" to a new page (child of SSO SAML).

arangograph/single-sign-on.md
nerpaula: Yes, let's instruct them to always share these via an ArangoGraph Support ticket.

arangograph/single-sign-on.md
nerpaula: Can you please provide a working example? not sure what needs to be substituted. If my organization ID is 123456, what value do I need to fill in?

arangograph/single-sign-on.md
ansoboleva: Agree, right now it all reads like it's required. @nerpaula let's move the SCIM topic out of the main content here.

arangograph/single-sign-on.md
```diff
@@ -25,17 +44,98 @@ can sync the information about user resources in real-time:
 - `GET /Users` - Lists all users.
```
ansoboleva, line 44: Lists --> List (consistency with other bullet points)

ansoboleva: in real-time --> in real time

arangograph/single-sign-on.md
```diff
@@ -25,17 +44,98 @@ can sync the information about user resources in real-time:
...
+[Okta Documentation](https://help.okta.com/en-us/Content/index.htm?cshid=csh-index){:target="_blank"}.
+
+### Create the SAML app integration in Okta
+
+1. Go to Okta and select **Applications** from the left sidebar menu.
+2. Click **Create App Integration**.
+3. In the **Create a new app integration** dialog, select **SAML 2.0**.
+
+   ![ArangoGraph Create Okta App Integration](images/arangograph-okta-create-integration.png)
+4. In the **General Settings**, specify a name for your integration.
```
ansoboleva, line 76: and click Next.
To avoid writing it separately as step 5.

arangograph/single-sign-on.md
```diff
@@ -25,17 +44,98 @@ can sync the information about user resources in real-time:
...
+   ![ArangoGraph Okta Integration Name](images/arangograph-okta-integration-name.png)
+5. Click **Next**.  
+6. Configure the SAML settings:
+   - For **Single sign-on URL**, use `https://auth.arangodb.com/login/callback?connection=ORG_ID`
+   - For **Audience URI (SP Entity ID)**, use `urn:auth0:arangodb:ORG_ID`
+   
+   {% hint 'info' %}
+   Replace **ORG_ID** with your organization identifier from the 
+   ArangoGraph Dashboard. To find your organization ID, go to the **User Toolbar**
+   in the top right corner, which is accessible from every view of the Dashboard,
```
ansoboleva, line 87: feels to me it can be helpful to provide a screenshot here

arangograph/single-sign-on.md
```diff
@@ -25,17 +44,98 @@ can sync the information about user resources in real-time:
...
+   and click **My organizations**.
+   {% endhint %}
+
+   ![ArangoGraph Okta SAML General Settings](images/arangograph-okta-saml-general-settings.png)
+7. In the **Attribute Statements** section, define the following custom attributes:
+   - email: `user.email`
+   - first_name: `user.firstName`
+   - last_name: `user.lastName`
+   - email_verified: `true`
+   - picture: `user.profileUrl`
```
ansoboleva, line 97: can we also add example values here?

arangograph/single-sign-on.md
```diff
@@ -25,17 +44,98 @@ can sync the information about user resources in real-time:
...
+
+   ![ArangoGraph Okta SAML Attributes](images/arangograph-okta-saml-attributes.png)
+8. Click **Next**.
+9. In the **Configure feedback** section, select **I'm an Okta customer adding an internal app**.
+10. Click **Finish**. The SAML app integration is now created.
+
+### SAML Setup
+
+1. After the integration is created, the **SAML Signing Certificates** section
```
ansoboleva, line 106: This isn't really a procedure. Can you please rewrite it a paragraph style with adding steps for the user actions only? And it's very important to provide concrete info on how to copy/share settings and certificate.

arangograph/single-sign-on.md
```diff
@@ -25,17 +44,98 @@ can sync the information about user resources in real-time:
...
+
+   ![ArangoGraph Okta SAML Setup](images/arangograph-okta-saml-setup.png)
+4. The setup instructions include the following items:
+   - Identity Provider Single Sign-On URL
+   - Identity Provider Issuer
+   - X.509 Certificate
+5. Copy the IdP settings, download the certificate, and share them with the
+   ArangoGraph team in order to complete the SSO configuration.
+
+### Enable SCIM provisioning in Okta
```
ansoboleva, line 119: Adding a couple sentences on when/why do it would be helpful.

ansoboleva (changes requested):

arangograph/scim.md
```diff
@@ -0,0 +1,78 @@
...
+
+ArangoGraph provides support to control and manage members access in
+ArangoGraph organizations with the
+**System for Cross-domain Identity Management** (SCIM) provisioning. 
+This enables you to propagate to ArangoGraph any user access changes by using
+the dedicated API.
+
+{% hint 'info' %}
+To enable the SCIM feature, get in touch with the ArangoGraph team via an
+ArangoGraph Support Ticket.
```
lakhansamani, line 20: Would be good to have link to Support page, here.

arangograph/scim.md
```diff
@@ -0,0 +1,78 @@
...
+- `GET /Users` - List all users.
+- `GET /Users/:user_id` - Get details for a given user ID.
+- `POST /Users` - Invite a new user to ArangoGraph.
+- `PUT /Users/:user_id` - Update a given user ID.
+- `DELETE /Users/:user_id` - Delete a specified user ID.
+
+ArangoGraph organization administrators can generate an API key for a specific organization.
+The API token consists of a key and a secret. Using this key and secret as the
+Basic Authentication Header (Basic Auth) in SCIM provisioning, you can access the APIs and
+manage the user resources.
```
lakhansamani, line 43: It will be good to mention here that don't set time to live here else you will have to keep rotating basic auth credentials in okta.
Screenshot 2023-02-14 at 4 56 08 PM

arangograph/single-sign-on.md
nerpaula: @ansoboleva this is basically a mapping between the ArangoGraph attribute names and Okta attributes. After defining these attributes, Okta automatically fills in the values for these attributes (picked from the Okta users list). 
Users need to define these as seen in the screenshot. Will rephrase for clarity.

arangograph/single-sign-on.md
```diff
@@ -9,33 +9,96 @@ title: Single Sign-On (SSO)
...
+   - **X.509 Certificate**
+4. Copy the IdP settings, download the certificate using the
+   **Download X.509 certificate** button, and share them with the ArangoGraph
+   team via an ArangoGraph Support Ticket in order to complete the SSO
+   configuration.
 
-To learn how to generate a new API key, see the [API Keys](my-account.html#api-keys) section.
+   {% hint 'warning' %}
+   Do not send this information over unencrypted channels, e.g. via normal email.   
+   {% endhint %}
```
mattcarabine, line 99: @Simran-B I know you requested to add it but could I ask that this is removed?

The reasoning is that none of what they'll be sharing with us is "sensitive", the signing certificate is the public certificate. Also, when raising a support ticket they'll be sent over email anyway, so I don't want to give the users the impression that this is "unsafe".

mattcarabine (changes requested):
Have requested one small change, but this looks so much better @nerpaula, flows very nicely, thank you very much!

arangograph/scim.md
nerpaula: since we don't have a generic link, rephrased this and pointed to the Request Help form.

arangograph/single-sign-on.md
nerpaula: Agreed with Simran that we can remove this warning.

mattcarabine (approved):

PR #1271, opened 2023-02-14
DOC-370 | ArangoGraph - certificates info and encryption

nerpaula:


arangograph/projects.md
```diff
@@ -115,9 +121,14 @@ certificate generation, both the default and high port serve the same
 self-signed certificate.
 {% endhint %}
 
-When a private endpoint is being used, you can specify alternate DNS names,
-which are added to the self-signed certificate only as Subject Alternative
-Name (SAN).
+When using private endpoints, notebooks, and metrics, you can specify
```
Robert-Stam, line 124: can you add a note in the PE section, so a customer can find this info more easily?

Robert-Stam (approved):
See remark, further LGTM

arangograph/projects.md
nerpaula: done

3.10/security-encryption.md
```diff
@@ -28,6 +28,11 @@ performance and resistance to side-channel attacks.
 
 The encryption feature is supported by all ArangoDB deployment modes.
 
+{% hint 'info' %}
+The ArangoGraph Insights Platform has encryption set on by default and cannot
```
Simran-B, line 32: Maybe mention that this refers to encryption at rest (encrypted on disk) as well as in transit (encrypted communication)?

nerpaula: yes, added it.

PR #1274, opened 2023-02-15
update examples for more clarity, fixed typos, and missing flags

whoiscarlo:
I updated the manual start cluster examples to include the --agency.endpoint flag that was missing on the creation of the agent instances and I updated the docker cluster example to be more clear and concise.


whoiscarlo:
I have already signed the CLA Agreement and emailed it to cla@arangodb.com. My resent it in because my github account wasn't attached. Let me know if there's anything I need to do.

fceller:
@mattcarabine @Simran-B CLA has been received

3.10/deployment-cluster-manual-start.md
```diff
@@ -310,25 +322,45 @@ arangod --server.authentication=false \
...
+  --server.authentication false \
+  --agency.activate true \
+  --agency.size 4 \
+  --agency.endpoint tcp://192.168.1.1:8531 \
+  --agency.endpoint tcp://192.168.1.2:8531 \
+  --agency.endpoint tcp://192.168.1.3:8531 \
+  --agency.endpoint tcp://192.168.1.5:8531 \
+  --agency.supervision true \
+  --database.directory agent
 ```
```
neunhoef, line 355: This is not true. We do not support changing the number of agents after deployment. This is going to cause serious trouble down the road, since the agency size must be constant (and equal) across all machines. Please remove this suggestion "Note 3".

3.10/deployment-cluster-manual-start.md
```diff
@@ -181,7 +187,10 @@ arangod --server.endpoint tcp://0.0.0.0:8531 \
   --agency.activate true \
   --agency.size 3 \
   --agency.supervision true \
-  --database.directory agent 
+  --agency.endpoint tcp://192.168.1.1:8531 \
+  --agency.endpoint tcp://192.168.1.2:8531 \
+  --agency.endpoint tcp://192.168.1.3:8531 \
```
neunhoef, line 192: Strictly speaking, this is not necessary, AFAIK. The first agent can be started without knowing others.

3.10/deployment-cluster-manual-start.md
```diff
@@ -193,6 +202,9 @@ arangod --server.endpoint tcp://0.0.0.0:8531 \
   --agency.activate true \
   --agency.size 3 \
   --agency.supervision true \
+  --agency.endpoint tcp://192.168.1.1:8531 \
+  --agency.endpoint tcp://192.168.1.2:8531 \
+  --agency.endpoint tcp://192.168.1.3:8531 \
```
neunhoef, line 207: AFAIK only the first agent would have to be given on the command line.

neunhoef (changes requested):
Thanks for the contribution. I think Note 3 is wrong, we do not support changing the number of agents after deployment. This is in fact dangerous and we must remove this Note again.
See comments.

3.10/deployment-cluster-manual-start.md
whoiscarlo: Ah yea. I learned that the hard way this week. So then what do you do if you need to increase the size of the agency or switch an exist agent to a new machine/ different IP address?

3.10/deployment-cluster-manual-start.md
whoiscarlo: K I'll update that.

3.10/deployment-cluster-manual-start.md
whoiscarlo: Alright cool, I'll update this now

PR #1275, opened 2023-02-17
ArangoSearch - Achieve stable pagination

nerpaula:


3.10/arangosearch-ranking.md
```diff
@@ -53,6 +53,29 @@ Scoring functions cannot be used outside of `SEARCH` operations, as the scores
...
 inverse document frequency (IDF).
 
+### Stable Pagination Results
+
+The SORT operation does not guarantee a stable sort if there is no unique value
+to sort by. This leads to an undefined order when sorting equal documents.
+
+To achieve stable pagination, you must meet the following requirements: 
+- the Dataset should not be changed
+- the SORT operation must have at least one unique field to sort by
```
Simran-B, line 63:
```suggestion
- the `SORT` operation must have at least one field with a unique value to sort by
```

3.10/arangosearch-ranking.md
```diff
@@ -53,6 +53,29 @@ Scoring functions cannot be used outside of `SEARCH` operations, as the scores
...
+
+When stable sort is required, you can use a tiebreaker field. If the application
+has a preferred field that indicates the order of documents with the same score,
+then this field should be used in the SORT operation as a tiebreaker. If there
```
Simran-B, line 67:
```suggestion
then this field should be used in the `SORT` operation as a tiebreaker. If there
```

3.10/arangosearch-ranking.md
```diff
@@ -53,6 +53,29 @@ Scoring functions cannot be used outside of `SEARCH` operations, as the scores
...
+is no such field, you can use the `_id` system field as it is unique and
```
Simran-B, line 68:
```suggestion
is no such field, you can use the `_id` system attribute as it is unique and
```

3.10/arangosearch-ranking.md
```diff
@@ -53,6 +53,29 @@ Scoring functions cannot be used outside of `SEARCH` operations, as the scores
...
+present in every document. 
+
+```aql
+FOR doc IN viewName
+  SEARCH ANALYZER(STARTS_WITH(doc._fulltext, 'movie name'), 'text_en')
+  SORT BM25(doc) DESC, doc._id
+  LIMIT 10,20
```
Simran-B, line 75:
```suggestion
  LIMIT 10, 20
```

Simran-B (changes requested):

Dronplane (approved):
LGTM. But I'd suggest to have a more detailed example with dataset,

nerpaula:
Added an example with dataset, thanks @Simran-B for helping out with this.

3.10/arangosearch-ranking.md
```diff
@@ -53,6 +53,67 @@ Scoring functions cannot be used outside of `SEARCH` operations, as the scores
...
 inverse document frequency (IDF).
 
+### Stable pagination for results
+
+The `SORT` operation does not guarantee a stable sort if there is no unique value
+to sort by. This leads to an undefined order when sorting equal documents.
+
+To achieve stable pagination, you must meet the following requirements: 
+- the dataset should not be changed
+- the `SORT` operation must have at least one field with an unique value to sort by
```
Simran-B, line 63: Sentence style?
```suggestion
- The dataset should not be changed.
- The `SORT` operation must have at least one field with an unique value to sort by.
```
Should we point out that even if the dataset doesn't change that you can get an unstable order with pagination? It's not so clear what the first point is supposed to say.

3.10/arangosearch-ranking.md
```diff
@@ -53,6 +53,67 @@ Scoring functions cannot be used outside of `SEARCH` operations, as the scores
...
+then this field should be used in the `SORT` operation as a tiebreaker. If there
+is no such field, you can use the `_id` system attribute as it is unique and
+present in every document. 
+
+```aql
+FOR doc IN viewName
+  SEARCH ANALYZER(STARTS_WITH(doc._fulltext, 'movie name'), 'text_en')
+  SORT BM25(doc) DESC, doc._id
+  LIMIT 10, 20
+  RETURN doc
```
Simran-B, line 76: I don't like that we start with this "free-floating" example and then switch to another one. I think we should show the full final example here instead, and possibly add a comment after `doc._id` to emphasize it.

3.10/arangosearch-ranking.md
```diff
@@ -53,6 +53,67 @@ Scoring functions cannot be used outside of `SEARCH` operations, as the scores
...
+
+| title | score |
+|:------|:-------|
+| Red Shadow: Akakage | 8.882122039794922 |
+| Beverly Hills Ninja | 7.128915786743164 | 
+| Naruto the Movie: Ninja Clash in the Land of Snow | 7.041049957275391 | 
+| TMNT | 7.002012729644775 | 
+| Batman Begins | 6.30634880065918 | 
+
+If you change the query to `LIMIT 5, 5`, then you get:
```
Simran-B, line 100:
```suggestion
If you change the query to `LIMIT 5, 5` to get the next few results, then you get:
```

3.10/arangosearch-ranking.md
```diff
@@ -53,6 +53,67 @@ Scoring functions cannot be used outside of `SEARCH` operations, as the scores
...
+| title | score |
+|:------|:-------|
+| Batman Begins | 6.30634880065918 | 
+| Shogun Assassin  | 5.8539886474609375 | 
+| Scooby-Doo and the Samurai Sword | 5.736422538757324 | 
+| Revenge of the Ninja | 5.212964057922363 | 
+| Winners & Sinners 2: My Lucky Stars | 5.165824890136719 | 
+
+Note how the position of Batman Begins is changed. It has the same score as
+Teenage Mutant Ninja Turtles II: The Secret of the Ooze.
```
Simran-B, line 111: We did not see the TMNT result with the score so far! And I would also mention that Batman Begins appears twice, which is the result of the unstable sort, but it wouldn't hurt to also explain the outcome from a user PoV.

Simran-B (approved):

PR #1276, opened 2023-02-17
Recommend Docker images on Windows

nerpaula:


3.10/installation-windows.md
```diff
@@ -12,6 +12,7 @@ You can install ArangoDB on Windows (64-bit) using different methods:
   - [attended](#installing-using-the-installer) (GUI)
   - [unattended](#unattended-installation-using-the-installer) (command line)
 - Manually, using a [ZIP archive](#installing-using-the-zip-archive)
+- Using a [Docker image](#docker)
```
Simran-B, line 15: > You can install ArangoDB on Windows (64-bit) using different methods:

This should be reworded as you don't really "install" Docker images ("You can use..."), and since Docker will be recommended way of using ArangoDB on Windows, I would put Docker images at the top of the list (which then also reflects the order of headlines)

3.10/installation-windows.md
```diff
@@ -24,6 +25,17 @@ or `certutil -hashfile <filename> sha256` in a terminal.
 Running production environments on Windows is not supported.
 {% endhint %}
 
+Docker
+------
+
+The recommended way of using ArangoDB on Windows is to use the
+[ArangoDB Docker images](https://www.arangodb.com/download-major/docker/){:target="_blank"}
```
Simran-B, line 32: This seems like an unnecessary indirection. I think we should link directly to Docker Hub.
I believe Vadim would like us to point out the difference between `arangodb` and `arangodb/arangodb` (they are not equal, `arangodb` are the official ones approved by Docker staff, whereas `arangodb/arangodb` are more up-to-date an published by us directly.

Simran-B (changes requested):

3.10/installation-windows.md
nerpaula: Added a link to `arangodb/arangodb`Docker images. Other details will be described in DOC-339 on a separate "Install with Docker" page.

3.10/installation-windows.md
```diff
@@ -17,19 +18,30 @@ Visit the official [Download](https://www.arangodb.com/download){:target="_blank
...
 {% hint 'info' %}
 Running production environments on Windows is not supported.
 {% endhint %}
 
+Docker
+------
+
+The recommended way of using ArangoDB on Windows is to use the
+[ArangoDB Official Docker images](https://hub.docker.com/_/arangodb){:target="_blank"}
+with, for instance, [Docker Desktop](https://www.docker.com/products/docker-desktop/){:target="_blank"}. You can also use the [Docker images](https://hub.docker.com/r/arangodb/arangodb){:target="_blank"} that are directly published by ArangoDB on a regular basis.
```
Simran-B, line 33: I think it would be clearer if we mentioned the image names, `arangodb` and `arangodb/arangodb` here, so that it's easier to distinguish the links.

PR #1280, opened 2023-02-20
Move log levels page to Administration, remove Monitoring section

Simran-B:


nerpaula (commented):
> See an example of how to configure log levels in the Administration chapter

Let's reword this now that we moved the page to the Administration chapter as well.

Simran-B:
Good catch. I reworked the content.

nerpaula (approved):

PR #1300, opened 2023-03-15
DOC-374 | Release notes for APM-73 (status code for write concern failure)

Simran-B:
Upstream: https://github.com/arangodb/arangodb/pull/18233/


Simran-B:
Needs clarification:

> If the command line option `--cluster.failed-write-concern-error-code=503` is set, then HTTP 503 is returned. Note that no cluster-internal retry is happening, such that a client is informed right away about the problem. Retry loops have to be organized in the client program.

versus

> The new, more correct behavior is to send a 503 SERVICE UNAVAILABLE, which leads to retries on the coordinator.

nerpaula:
https://jenkins.arangodb.biz:3456/view/Documentation/job/arangodb-ANY-examples/2963/

nerpaula (approved):

PR #1221, opened 2022-12-13
[PREG-183] Remove async mode from pregel

jvolmer:
Deletes async code - including AsyncSCC and related code. It is not tested and the code is hard to follow with it because it is hacked in.

(Code PR: https://github.com/arangodb/arangodb/pull/17802)

PR #1301, opened 2023-03-16
Deprecation/Removal notice of two Pregel features for  3.11

markuspf:
This PR adds deprecation/removal notices for two Pregel features, going into 3.11:
1. The 'Custom Pregel' has been removed.
2. The DMID algorithm has been deprecated.

Related issues:
1. https://arangodb.atlassian.net/browse/GORDO-1513
2. https://arangodb.atlassian.net/browse/GORDO-1588

Since we're still unfamiliar with the docs repo, we're not sure if we've made all the necessary edits and in the right places. Please let us know in the review.


3.11/appendix-deprecated.md
```diff
@@ -32,6 +32,12 @@ considered obsolete and may get removed in a future release. They are currently
 kept for backward compatibility. There are usually better alternatives to
 replace the old features with:
 
+- **Custom Pregel**:
+  The experimental feature "Custom Pregel" has been removed.
```
adityamukho, line 36: Not sure if this removal notice belongs to the deprecated features list. For example, I didn't see a removal notice for the async feature being added to this file - ref: https://github.com/arangodb/docs/pull/1221/files

@Simran-B please clarify.

Simran-B: It's not intended as a changelog, but rather as a page we can redirect removed URLs to. There are no hard rules but the convention is to add larger features here when they are deprecated/removed.

PR #1315, opened 2023-03-28
Limitations of Spark speculative execution

rashtao:


drivers/spark-connector-new.md
```diff
@@ -281,6 +279,7 @@ When writing to an edge collection (`table.type=edge`), the schema of the Datafr
 - Writing records with the `_key` attribute is only allowed on collections sharded by `_key`. 
 - In case of the `Append` save mode, failed jobs cannot be rolled back and the underlying data source may require manual cleanup.
 - Speculative execution of tasks would only work for idempotent write configurations. See [Write Resiliency](#write-resiliency) for more details.
+- Speculative execution of tasks could cause concurrent writes of same documents, resulting in write-write conflicts or lock timeouts
```
Simran-B, line 282: @rashtao Is there speculative execution? We should not use subjunctives in that case.
```suggestion
- Speculative execution of tasks only works for idempotent write configurations. See [Write Resiliency](#write-resiliency) for more details.
- Speculative execution of tasks can cause concurrent writes to the same documents, resulting in write-write conflicts or lock timeouts
```

Simran-B (approved):

PR #1316, opened 2023-03-29
DOC-385 | Add overview of Tools to Getting Started

nerpaula:
preview https://deploy-preview-1316--zealous-morse-14392b.netlify.app/docs/3.10/quick-start-guide.html


3.10/quick-start-guide.md
```diff
@@ -21,6 +22,65 @@ update, replace, and remove documents using ArangoDB's web interface or a driver
 
 Learn more about ArangoDB's [data model and concepts](data-model-and-concepts.html).
 
+## Available Tools in ArangoDB
+
+ArangoDB provides various tools to help you manage and work with your data.
+
+- [ArangoDB Server](programs-arangod.html) is the core component of ArangoDB that
+  stores data and handles requets. You can communicate with the server through
+  the web interface, command-line interface, drivers, and REST API.
```
Simran-B, line 31: I think it's important to mention the name of the executable (arangod) and clarify the difference between an ArangoDB server as in an instance of the database system (which can be comprised of multiple arangod processes), the server binary, and a server node. For clusters, you shouldn't communicate with any server node but only Coordinators - not something that necessarily needs to be mentioned here, but perhaps there is a wording that leaves this room for interpretation.

3.10/quick-start-guide.md
Simran-B, line 1: We have a dedicated page for "How to interact with ArangoDB" and this content seems like it would fit there. For the quick start guide, I would expect something shorter, for example, leaving out arangobench, and not focusing on the different executables/tools, but rather creating groups by topic. What I mean is that it should look at the tools more from a user perspective and be worded accordingly, like "get data into and out of ArangoDB" as a topic, covering arangoimport + arangoexport, but also mentioning other options like the API and the JSON upload/download option in the web interface (which is pretty limited but still). Similarly, arangodump, arangorestore, and arangobackup can be grouped together under a topic for backing up and restoring data/the system.

Simran-B (changes requested):

3.10/quick-start-guide.md
```diff
@@ -13,11 +13,14 @@ This chapter introduces ArangoDB's core concepts and covers the following:
...
-see [How to Interact With ArangoDB](how-to-interact-with-arangodb.html).
-For example, you can create and drop databases/collections as well as save,
-update, replace, and remove documents using ArangoDB's web interface or a driver.
+using ArangoDB's interactive command-line interface called [arangosh](programs-arangosh.html).
+
+See [How to Interact With ArangoDB](how-to-interact-with-arangodb.html) to learn
+more about:
+- Ways you can use and communicate with ArangoDB servers
+- How to set up and deploy an ArangoDB instance
+- Get data in and out of ArangoDB
```
Simran-B, line 22: in or into?

Simran-B (changes requested):

3.10/how-to-interact-with-arangodb.md
```diff
@@ -62,6 +63,54 @@ may use the low-level REST API directly as needed.
...
 
+## Set Up and Deploy ArangoDB
+
+[ArangoDB Starter](programs-starter.html) (`arangodb` binary) helps you set up
+and deploy ArangoDB instances on bare-metal servers and supports all ArangoDB
+deployment modes, such as a single server instance, Active Failover, and cluster
+(including Datacenter-to-Datacenter Replication).
+
+In addition to the Starter, there are also other ways that can be used to deploy
+ArangoDB:
```
Simran-B, line 74:
```suggestion
In addition to the Starter, there are also other ways that you can use to deploy
ArangoDB:
```

3.10/how-to-interact-with-arangodb.md
```diff
@@ -62,6 +63,54 @@ may use the low-level REST API directly as needed.
...
+- Running `arangod` directly
```
Simran-B, line 75:
```suggestion
- Run the `arangod` executable directly
```

3.10/how-to-interact-with-arangodb.md
```diff
@@ -62,6 +63,54 @@ may use the low-level REST API directly as needed.
...
+- [Docker containers](install-with-docker.html)
+- [Kubernetes](deployment-kubernetes.html)
+- Using installation packages
```
Simran-B, line 78:
```suggestion
- Use installation packages
```

PR #1326, opened 2023-04-11
DOC-420 | High level restructuring

nerpaula:
To do:

- [x] decide the exact order of top level menu entries and subtitles
- [x] apply changes to 3.11
- [x] apply changes to 3.12

Considering that the migration script takes all `subtitles` found in the yaml file and turns them into expandable sections in hugo, we may need to go for a flat structure here. For example, see the Graphs chapters where all subpages are now top level; we need this flat structure so that we don't get unnecessary deep nesting in hugo. Same needs to be applied to other chapters as well.


_data/3.11-manual.yml
```diff
@@ -1,186 +1,145 @@
...
+        - text: Computed Values
+          href: data-modeling-documents-computed-values.html
+    - text: Views
+      href: data-modeling-views.html
+      children:
+        - text: Database Methods
+          href: data-modeling-views-database-methods.html
+        - text: View Methods
+          href: data-modeling-views-view-methods.html
+- text: Operational Factors
```
Simran-B, line 61: This is actually about data modeling (as in architecting solutions), whereas the above pages are about the data model of ArangoDB. We had both mixed together for a very long time and eventually separated the topics because they are quite different. Now, they are mixed together again. Is this really what we want?

_data/3.11-manual.yml
```diff
@@ -207,119 +166,47 @@
...
+    - text: Ranking
+      href: arangosearch-ranking.html
+    - text: Performance
+      href: arangosearch-performance.html
+    - text: <code>search-alias</code> Views Reference
+      href: arangosearch-views-search-alias.html
+    - text: <code>arangosearch</code> Views Reference
+      href: arangosearch-views.html
+- subtitle: TRANSFORM DATA 
+- text: Analyzers
```
Simran-B, line 207: Not a fan of having a separate section about transforming data but only one chapter

_data/3.11-manual.yml
```diff
@@ -644,6 +441,212 @@
       children:
         - text: Agency Dump
           href: troubleshooting-cluster-agency-dump.html
+- subtitle: TOOLS
+- text: Programs & Tools
```
Simran-B, line 445: Another section with only one chapter in it, and it's not just tools. Would calling it "ecosystem" or something broader make sense? Or split it into Server and Tools?

_data/3.11-manual.yml
nerpaula: We need to find better names for these two, more distinguishable. How about something like:
- Data Model & Concepts -> How data is structured / Data structure in ArangoDB
- Data Modeling & Operational Factors > Model application data for ArangoDB

Agree that we should keep them separate. The question is where to place the second one (that's why it ended up here)?

_data/3.11-manual.yml
nerpaula: Indeed, not desirable but we don't have other content. A possible solution is to move it under "Index, search, transform data". We need to check if the subtitle is not too lengthy though.

_data/3.11-manual.yml
nerpaula: Yes, let's split it into Server and Tools. "Ecosystem" sounds good as a subtitle.

_data/3.11-manual.yml
```diff
@@ -441,101 +328,11 @@
...
-            - text: GraphQL
-              href: foxx-reference-modules-graph-ql.html
-            - text: Queues
-              href: foxx-reference-modules-queues.html
-    - text: Deployment
-      href: foxx-deployment.html
 - text: SatelliteCollections
   href: satellites.html
 - text: SmartJoins
   href: smartjoins.html
```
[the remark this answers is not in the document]
nerpaula: Yes, these topics that fit nowhere... Does it make sense to place them under Replication chapter? Or somewhere close to Cluster? If not, the only option I see is in Collections...

_data/3.11-manual.yml
Simran-B: "Data model" and "Data modeling" are indeed too similar. Maybe something along the line of _planning_ and _designing_ solutions with ArangoDB? It could become the place for how to model application data and architecting solutions, although it would currently be pretty empty as we don't have much content.

_data/3.11-manual.yml
Simran-B: While Analyzers are a standalone feature, they are most often used in combination with ArangoSearch. We shouldn't treat it like a component of ArangoSearch as you can use it independently, but placing it next to the indexing and search content would be okay in my opinion. Finding a succinct title is hard, though.

_data/3.11-manual.yml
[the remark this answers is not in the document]
Simran-B: They are features that you as a solution architect or developer need to plan for to take advantage of the performance benefits, so "modeling data" / "plan and design solutions" would be fitting. We could also consider adding a chapter about scalability features, but would we move the graph-related features to that place instead of having them under Graphs?

_data/3.10-manual.yml
```diff
@@ -1,188 +1,145 @@
-- subtitle: GETTING STARTED
+- subtitle: INTRODUCTION
 - text: About ArangoDB
   href: index.html
   children:
```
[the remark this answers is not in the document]
Simran-B: Seems accidental. The content that used to be under About ArangoDB has moved up one level.

ansoboleva (approved):

PR #1327, opened 2023-04-11
DOC-354 | APM-407 | AQL bulk INSERT operations

Simran-B:
Julia says:
> it reduces the amount of http requests, but I'm not sure it reduces setup and shutdown

Maybe I misunderstand how it achieves better performance
. I thought it does no longer need to contact all DB-Servers first but only send the relevant portion of the data to the right DB-Server.

What happens **without** this optimization?
 There is batching, but do the entire input documents also get sent to all DB-Servers instead of just a fraction of them? That's how it looks like in the query explain output.


cpjulia (approved):
LGTM

3.11/aql/execution-and-performance-optimizer.md
```diff
@@ -717,6 +721,36 @@ The following optimizer rules may appear in the `rules` attribute of
   case, no AQL is executed on the DB-Servers. Instead, the Coordinator
   directly works with the documents on the DB-Servers.
 
+- `optimize-cluster-multiple-document-operations`:
+  Bulk `INSERT` operations in cluster deployments can be optimized to avoid
+  unnecessary overhead that AQL queries typically require for the setup and
+  shutdown in clusters.
```
mpoeter, line 727:
```suggestion
  Bulk `INSERT` operations in cluster deployments can be optimized to avoid
  unnecessary overhead that AQL queries typically require for the setup and
  shutdown in clusters, as well as for the internal batching.
```

3.11/aql/execution-and-performance-optimizer.md
```diff
@@ -717,6 +721,36 @@ The following optimizer rules may appear in the `rules` attribute of
...
+
+  - There are no following `RETURN` nodes (including any `RETURN OLD` or `RETURN NEW`)
+  - The `FOR` loop is not contained in another outer `FOR` loop or subquery
+  - There are no other operations (e.g. `LET`, `FILTER`) between `FOR` and `INSERT`
+  - `INSERT` is not used on a SmartGraph edge collection
+  - The `FOR` loop iterates over a constant, deterministic expression
+
+  The optimization then adds a `MultipleRemoteExecutionNode` to the query
+  execution plan, which takes care of inserting all documents into the collection
+  in one go. Further optimizer rules are skipped if the optimization is triggered.
```
mpoeter, line 752:
```suggestion
```suggestion
  The optimization then replaces the `InsertNode` and `EnumerateListNode` with a `MultipleRemoteExecutionNode` in the query
  execution plan, which takes care of inserting all documents into the collection
  in one go. Further optimizer rules are skipped if the optimization is triggered.
```

mpoeter:
Without the new rule a query like this `FOR d in [{}] INSERT d INTO col` results in the following execution plan:
```
Execution plan:
 Id   NodeType            Site  Est.   Comment
  1   SingletonNode       COOR     1   * ROOT
  2   CalculationNode     COOR     1     - LET #2 = [ {  } ]   /* json expression */   /* const assignment */
  3   EnumerateListNode   COOR     1     - FOR d IN #2   /* list iteration */
  9   CalculationNode     COOR     1       - LET #4 = MAKE_DISTRIBUTE_INPUT_WITH_KEY_CREATION(d, null, { "allowSpecifiedKeys" : false, "ignoreErrors" : false, "collection" : "col" })   /* simple expression */
  5   DistributeNode      COOR     1       - DISTRIBUTE #4
  6   RemoteNode          DBS      1       - REMOTE
  4   InsertNode          DBS      0       - INSERT #4 IN col 
  7   RemoteNode          COOR     0       - REMOTE
  8   GatherNode          COOR     0       - GATHER   /* parallel, unsorted */
```
Here the coordinator requests data from the DB servers (the GatherNode), but the DB servers in turn have to go back to the coordinator in order to request their data (the DistributeNode). In addition we also have the AQL setup and finish requests. Also, we usually have batches of 1000 documents. In case the input array contains several thousand documents, this variant has to process multiple batches including the corresponding network requests.

With the new rule enabled the execution plan looks as follows:
```
Execution plan:
 Id   NodeType                         Site  Est.   Comment
  1   SingletonNode                    COOR     1   * ROOT
  2   CalculationNode                  COOR     1     - LET #2 = [ {  } ]   /* json expression */   /* const assignment */
  5   MultipleRemoteModificationNode   COOR     1     - FOR d IN #2 INSERT d IN col
 ```
As you can see, the query now runs completely on the coordinator, so no AQL setup/finish requests are sent to db servers. Instead, the `MultipleRemoteModificationNode` starts an internal cluster transaction (El Cheapo) and performs a bulk document insert (basically the same as the normal document API) for _the whole input array in one go_. The cluster transaction still needs separate begin/commit requests, but according to our tests those have less overhead than the AQL setup/finish.

mpoeter (approved):

nerpaula (approved):

PR #1328, opened 2023-04-11
DOC-458 | MDS-1001 | Optimize access to last element in traversal

Simran-B:
Also adjust statement about number-suffixed optimizer rules and remove now obsolete rule from list


mchacki (approved):
LGTM

nerpaula (approved):

PR #1329, opened 2023-04-11
DOC-355 | Extended names for collections, Views, and indexes

Simran-B:
Related: https://github.com/arangodb/arangodb/pull/18697


3.11/aql/fundamentals-syntax.md
```diff
@@ -255,8 +255,23 @@ FOR doc IN `my-coll`
...
 operator for subtraction in AQL. The backticks escape the collection name to
-refer to the collection correctly. Note that quoting the name with `"` or `'`
-is not possible for collections.
+refer to the collection correctly.
+
+If you use extended collection and View names
+([`--database.extended-names` startup option](../programs-arangod-options.html#--databaseextended-names)),
+they may contain spaces, or non-ASCII characters such as Japanese or Arabic
+letters, emojis, letters with accentuation, and other Unicode characters.
+Escaping is required in these cases, too:
```
jsteemann, line 264: Nit: I don't think that _escaping_ is required here, but rather _quoting_.

3.11/data-modeling-collections.md
```diff
@@ -43,19 +43,66 @@ for the entire ArangoDB instance.
...
-
-- Names of user-defined collections must always start with a letter.
-  System collection names must start with an underscore. You should not use
-  system collection names for your own collections.
-
-- The maximum allowed length of a name is 256 bytes.
+{% hint 'info' %}
+The extended naming constraints are an **experimental** feature but they will
+become the norm in a future version. Drivers and client applications
+should be prepared for this feature.
```
jsteemann, line 56: The sentence "Drivers and client applications should be prepared for this feature" is ambiguous to me.
It can mean both:
1. we expect drivers & apps to be ready for this feature, so it is fine to use it.
2. we expect you to check if drivers & apps are ready, so use it with care.

I guess the intended meaning is that of 2.
So I suggest rephrasing as "Drivers and client applications need to be compatible with this feature. This should be ensured before turning the feature on." or such.

3.11/aql/fundamentals-syntax.md
Simran-B: Hm, I think it was already called escaping further up in the existing text. Need to check where else we might say escaping but mean quoting...

jsteemann: Definitely possible. But I think "quoting" would be the more correct term here.

nerpaula:
We should also mention that we support UTF-8 for collection, view, index names in the web interface.

jsteemann (approved):
LGTM

PR #1330, opened 2023-04-12
DOC-361 | Release notes for Custom Queries UI, new sort mechanism and search box

nerpaula:


Simran-B (commented):
Isn't the feature called "Saved queries"?
"Custom queries" can be confused with this startup option: https://www.arangodb.com/docs/stable/programs-arangoexport-options.html#--custom-query

nerpaula:
> Isn't the feature called "Saved queries"? "Custom queries" can be confused with this startup option: https://www.arangodb.com/docs/stable/programs-arangoexport-options.html#--custom-query

Looks like we call it "Custom queries" here: https://www.arangodb.com/docs/3.10/programs-web-interface-aql-editor.html#custom-queries

Simran-B:
Is it used anywhere else? I don't see it in the UI itself. There's the Queries toggle and the save buttons, hence the term "saved queries". It seems to be called both custom queries and saved queries in the CHANGELOG of the main repo (4 times vs. 3 times)...

PR #1336, opened 2023-04-15
add example combining POSITION and the array expansion operator

larsborn:
I initially made the mistake of doing something like this:

```
FOR row IN coll
   FOR entry in doc.array
      FILTER entry.field_name != @field_name
   LIMIT @limit
   RETURN DISTINCT doc
```

I now realize that this is _not_ how it works (the last two liens should be indented one more and hence this will in most cases not yield `@limit` many entries.

The intention behind this PR is to add a concrete example pushing folks in the right direction to do something like this

```
FOR row IN coll
    FILTER POSITION(doc.array[*].field_name , @field_name) == false
    LIMIT @limit
    RETURN doc
```

instead.


larsborn:
Sorry for the duplication, that was me sucking at git 🙈 .

PR #1337, opened 2023-04-15
Example for recursive fields

larsborn:
Submitting this PR per a discussion on Slack with @Dronplane and @Simran-B. When I first read this part of the documentation it wasn't clear to me where exactly to put the sub `fields` array.

### DIFF
--- a/3.10/aql/functions-array.md
+++ b/3.10/aql/functions-array.md
@@ -547,6 +547,10 @@ Return whether *search* is contained in *array*. Optionally return the position.
 To determine if or at which position a string occurs in another string, see the
 [CONTAINS() string function](functions-string.html#contains).
 
+If you want to search a list of objects,
+[the array expansion operator [*]](advanced-array-operators.html#array-expansion) might also be of
+interest. The third example below will illustrate its usage in this context.
+
 **Examples**
 
 {% aqlexample examplevar="examplevar" type="type" query="query" bind="bind" result="result" %}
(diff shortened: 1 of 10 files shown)


Simran-B:
@larsborn Should the 3rd POSITION() example point out that you don't need POSITION() if you only want to test whether a value is contained in a list of objects (boolean), like `FILTER "value" IN doc.arr[*].attr`?

larsborn:
I'll also make that point in the paragraph above. (in particular where `returnIndex` is described). Reasoning: When I read that sentence I thought `POSITION` is the way to go here. Along the lines of "why would there be the `returnIndex` argument if I'm supposed to use something entirely different?".

larsborn:
@Simran-B thanks for your patience and the additional paragraphs. This is ready for your review and thanks in advance!

Simran-B:
@larsborn Thanks! `returnIndex` is specifically about getting the index back, which isn't possible with the IN operator. It makes more sense to me to reference it in a different place. Given the existing content and conventions, it makes the most sense to me to place it below the parameter description.

larsborn:
Sounds good!

Simran-B:
A co-worker pointed out that using POSITION() can in fact be faster than using the IN operator, which I wasn't aware of. The function is compiled C++ code, whereas the AQL language construct iterates over VelocyPack data - which is potentially slower. However, you need to use the IN operator if you want to utilize indexes, and then it's a different story again. I thus reverted the recommendation of using IN over POSITION() that I had added.

Simran-B:
Our example generation doesn't support forks, but I tested it locally and will use https://github.com/arangodb/docs/pull/1340 to re-generate them now.

larsborn:
Thanks a lot!

> A co-worker pointed out that using POSITION() can in fact be faster than using the IN operator, which I wasn't aware of. The function is compiled C++ code, whereas the AQL language construct iterates over VelocyPack data - which is potentially slower. However, you need to use the IN operator if you want to utilize indexes, and then it's a different story again. I thus reverted the recommendation of using IN over POSITION() that I had added.

This is an interesting detail. Just so I fully understand it: given the list to iterate is very large then `POSITION` is (probably) faster if there's no index on the field being searched and `IN` is faster (probably) if there is an index? If that's really the case, should we document that? (happy to cobble together a first draft here again)

Simran-B:
Basically yes, but I don't have any data to back this up, and there could be additional aspects, like memory utilization. There has also been a recent change to avoid copying data that might be related to the IN operator, although I'm not sure, but there is always a chance that the AQL optimizer is improved to automatically transform queries to better execution plans. Therefore, I wouldn't currently add any recommendations to the docs without hard evidence that one approach is clearly better than the other for a given situation, across different hardware configurations.

PR #1343, opened 2023-04-20
added more infos to release notes

jsteemann:
These things were missing from release notes previously.


Simran-B (approved):
Thanks, Jan! I also added the features added after 3.10.0 to the 3.11 release notes

PR #1345, opened 2023-04-20
add requirement for system page size to documentation

jsteemann:
Inspired by https://arangodb.atlassian.net/browse/BTS-1163. 
The requirement for the operating system page size to be <= 4096 was not documented anywhere.
This PR adds it.


Simran-B (approved):
Thanks! Added the requirement to one more relevant page

PR #1348, opened 2023-04-25
clarify that unicodes names must be NFC-normalized

jsteemann:
Created to clarify that there is a requirement that database names, collection names, view names and index names with Unicode characters must be properly NFC-normalized, or they will be rejected by the server.


dothebart:
the current implementation state *normalizes* database names, but not collections, views, index names.

dothebart (approved):
With the latest changes database normalisation will be removed, hence LGTM.

Simran-B (approved):

PR #1358, opened 2023-05-05
DOC-352 | Auto login to database UI

nerpaula:
Added documentation for the auto login feature from ArangoGraph to database UI.


arangograph/deployments.md
```diff
@@ -176,24 +176,38 @@ a self-signed certificate is recommended. Read more on the
...
+### Auto login to database UI
+
+The auto login feature is a convenient and time-saving functionality that
+allows you to seamlessly access the ArangoDB database web interface from within
+the ArangoGraph platform without the need for additional login credentials.
+
+You can enable this feature in the **Password Settings** dialog. Please note
+that it may take a few minutes to get activated.
+Once enabled, you no longer have to fill in the `root` user and password of
+your ArangoDB deployment. This feature can be disabled at any time.
```
Robert-Stam, line 209: IMO we should describe something about:
1) permissions (currently we have full-access and read-only)
2) that it's available on port 8529 only (not 18529, which is for drivers). 

Limitations:
port 8529 will be technically switched from tcp to http mode, which means that velocy-stream cannot be used, so if drivers are (by accident) using port 8529 this will be a breaking change.

Robert-Stam (commented):
See comments, sorry for the delay (I was on holidays)

arangograph/deployments.md
nerpaula: Thanks, added this information.

mattcarabine (changes requested):

arangograph/deployments.md
```diff
@@ -176,24 +176,54 @@ a self-signed certificate is recommended. Read more on the
...
```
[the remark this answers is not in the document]
nerpaula: Added.

mattcarabine (approved):

PR #1374, opened 2023-05-19
Remove mention of `--ignore-distribute-shards-like-errors` startup option

Simran-B:
It's defunct in arangorestore for a long time and would likely cause problems anyway.

v3.12 only for now until reviewed. Should we still mention the working `--ignore-distribute-shards-like-errors` of arangodump?

The startup option of both arangodump and arangorestore will remain visible in the generated list of options until obsoleted/removed from the code. The same goes for arangodump's error message when you try to dump without the prototype collection (second sentence):

> ERROR [f7ff5] {dump} An error occurred: Collection clonedCollection's shard distribution is based on that of collection prototypeCollection, which is not dumped along. You may dump the collection regardless of the missing prototype collection by using the --ignore-distribute-shards-like-errors parameter.


mchacki (approved):
LGTM

nerpaula (approved):

PR #1394, opened 2023-06-05
DOC-479 | Release notes for 3.9.11 and 3.10.7

nerpaula:


nikita-vanyasin:
@nerpaula @Simran-B can you please add next Known Issue for v3.10.7?

This is new issue which will be introduced with the release of v3.10.7 and expected to be fixed with next patch release.

```
(EE only)
arangosync v2.18.0 contains an issue where synchronization can be interrupted with errors like "CloneDirectMQToken failed". This can happen after network disturbances on just after some time of running replication (6h>). Possible solutions:
1) use arangosync v2.17.0 binary instead. You can find it in previous packages of ArangoDB (e.g. v3.9.10).
2) stop synchronization and start again.
The issue will be fixed in v2.18.1 release of arangosync.
```

Simran-B (approved):

PR #1395, opened 2023-06-06
DOC-474 | Add `legacy` property for `geojson` Analyzers, 3.10 release notes

Simran-B:
Upstreams:
- devel: https://github.com/arangodb/arangodb/pull/19187
- 3.11: https://github.com/arangodb/arangodb/pull/19190
- 3.10: https://github.com/arangodb/arangodb/pull/19191


3.10/release-notes-upgrading-changes310.md
```diff
@@ -79,6 +79,43 @@ suddenly interpreted in a different way. See
...
+Analyzers. If these Analyzers are used in `arangosearch` Views, then they need
+to be dropped as well before dropping the Analyzers, and recreated after creating
+the new Analyzers.
+
+GeoJSON polygons are interpreted slightly differently (and more correctly)
+in the newer versions.
+
+Legacy `geojson` Analyzers continue to work and continue to produce the
+same results as in earlier versions, since they have the `legacy` property
+implicitly set to `true`.
```
MBkkt, line 95: It's not true, because in early 3.10 legacy flag wasn't implemented so it implicitly was created with false.
So it always created/opened with false by default.
If you need other behavior it's possible to recreate analyzer with force flag

MBkkt (changes requested):

3.10/release-notes-upgrading-changes310.md
Simran-B: Okay, so upgrading to 3.10+ means you get a different behavior with the Analyzer unless you manually enable `legacy`, whereas geo indexes have backward compatibility out of the box.

In what exact version has the `legacy` property been added?

What do you mean by `recreate analyzer with force flag`? Accordingly to Andrei, it is necessary to drop and recreate Views which use `geojson` Analyzers or you get inconsistencies.

MBkkt: > What do you mean by recreate analyzer with force flag

```
analyzer.remove(my_old_geojson_with_implicit_legacy_false_name, /*force=*/true);
analyzer.save(my_old_geojson_with_implicit_legacy_false_name, ...legacy=true)
```

MBkkt: > whereas geo indexes have backward compatibility out of the box.

Yes, when it was done, search tasks was just putted to the backlog...

> In what exact version has the legacy property been added?

3.10.5

Dronplane: Unfortunately there is no way to update analyzer in the link definition. So the only way would be to recreate the view IF there is polygons that require legacy behaviour.

nerpaula:
https://jenkins.arangodb.biz:3456/view/Documentation/job/arangodb-ANY-examples/3302/

nerpaula (approved):

PR #1398, opened 2023-06-10
Fix hint overflow & some geo content issues

Simran-B:
- `display: flex` makes the items default to `min-width: auto`, which made the content of the inner `<div>` overflow the hint box without showing a scrollbar. Setting `min-width: 0` makes the normal text wrap and displays a scrollbar for code blocks if it's too wide - affected page: https://www.arangodb.com/docs/stable/aql/functions-geo.html#within
- Don't substitute all line breaks after rendering the hint box Markdown body - this was implemented to avoid stray `</div>` from showing if a hint is placed inside of a Markdown list, which apparently closed elements prematurely. Line breaks are now preserved inside preformatted blocks to not put all of the code on a single line 
- Address some mistakes and formatting issues in the content about deprecated geo functions

### DIFF
--- a/3.10/aql/functions-geo.md
+++ b/3.10/aql/functions-geo.md
@@ -33,7 +33,7 @@ which is sufficient for most use cases such as location-aware services.
 DISTANCE(52.5163, 13.3777, 50.9322, 6.94) // 476918.89688380965 (~477km)
 
 // Sort a small number of documents based on distance to Central Park (New York)
-FOR doc IN doc // e.g. documents returned by a traversal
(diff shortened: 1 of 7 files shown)


Simran-B:
Fixed hint: https://deploy-preview-1398--zealous-morse-14392b.netlify.app/docs/3.10/aql/functions-geo.html#near

No regression here: https://deploy-preview-1398--zealous-morse-14392b.netlify.app/docs/3.10/arangograph/cloud-migration-tool.html#cloud-migration-workflow-for-minimal-downtime

nerpaula (approved):

PR #1407, opened 2023-06-15
Colon character should be avoided in document keys as it can conflict with EE features

Simran-B:


mchacki (approved):
LGTM.

I like the clarification thank you

PR #1425, opened 2023-07-06
APM-665 - overview of memory usage and options

neunhoef:
This PR will change the chapter on "Reduce memory footprint" to
give a brief overview over ArangoDB's memory usage and point towards
options to control the behaviour.

Later, we might add links to metrics for RAM usage visibility in that
chapter.

### DIFF
--- a/3.11/administration-reduce-memory-footprint.md
+++ b/3.11/administration-reduce-memory-footprint.md
@@ -302,13 +302,6 @@ other write operations, and can thus be limited there. When there are
 many or large open transactions, this can sum up to a large amount of
 RAM usage.
 
-Once a transaction is committed and data is supposed to written out to
-level 0 SST files, there is RAM usage for table builders.
-
-As mentioned above in the block cache section, RAM usage for table
-builders can be accounted for with the block cache, if
-`--rocksdb.reserve-table-builder-memory` is set to `true`.
-
 A further limit on RAM usage can be imposed by setting the option
 
 ```
--- a/3.11/administration-reduce-memory-footprint.md
+++ b/3.11/administration-reduce-memory-footprint.md
@@ -62,8 +62,10 @@ This will essentially scale down `arangod`'s memory usage to the
 given value. This is for example a first measure if more than
 one `arangod` server are running on the same machine.
 
-For more detailed tuning of the various subsystems, the sections below
-will be useful.
+Note however, that certain subsystems will then still be able to use
+an arbitrary amount of RAM, depending on the load from the user. If you
+want to protect your server against such misusages and for more detailed
+tuning of the various subsystems, the sections below will be useful.
 
 Before we get into the nitty-gritty details, we provide an overview over
 the different subsystems of ArangoDB that are using significant amounts of RAM.
--- a/3.11/administration-reduce-memory-footprint.md
+++ b/3.11/administration-reduce-memory-footprint.md
@@ -15,123 +15,436 @@ Do not apply any of the changes suggested here before you have tested them in
 in a development or staging environment.
 {% endhint %}
 
-ArangoDB's memory usage can be restricted and the CPU utilization be reduced
-by different configuration options:
+Usually, a database server will try to use all the memory it can get to
+improve performance by caching or buffering. Therefore it is important
+to tell an ArangoDB process how much memory it is allowed to use.
+ArangoDB detects the available RAM on the server and divides this up
+amongst its subsystems in some default way, which is suitable for a wide
+range of applications. However, there will be situations in which there
+are good reasons why these defaults are not suitable and the
+administrator will therefore want to fine tune memory usage. The two
+major reasons for this are:
 
-- storage engine
-- edge cache
-- server statistics
-- background threads
-- V8 (JavaScript features)
-- operating system / memory allocator (Linux)
+ - something else (potentially another `arangod` server) is running on
+   the same machine so that our `arangod` is supposed to use less than
+   the available RAM, and/or
+ - the actual usage scenario makes it necessary to increase the memory
+   allotted to some subsystem at the cost of another to achieve higher
+   performance for specific tasks.
 
-There are settings to make it run on systems with very limited resources, but
-they may also be interesting for your development machine if you want to make it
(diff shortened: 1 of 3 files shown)


3.11/administration-reduce-memory-footprint.md
```diff
@@ -15,123 +15,436 @@ Do not apply any of the changes suggested here before you have tested them in
...
+- RocksDB background compaction
+
+It is important to understand that all these have a high impact on
+performance. For example, the block cache is there such that already
+used blocks can be retained in RAM for later access. The larger the
+cache, the more blocks can be retained and so the higher the probability
+that a subsequent access does not need to reach out to disk and thus
+incurs no additional costs on performance and IO capacity.
+
+RocksDB caches bloom filters and index blocks of its SST files in RAM.
```
jsteemann, line 83: As this is controled via configuration, I would make this conditional:
```suggestion
RocksDB can cache bloom filters and index blocks of its SST files in RAM.
```

3.11/administration-reduce-memory-footprint.md
```diff
@@ -15,123 +15,436 @@ Do not apply any of the changes suggested here before you have tested them in
...
+(set to unlimited by default!):
+
+```
+--rocksdb.total-write-buffer-size
+```
 
-Under normal write load, the write buffers will use less than 1 GByte of memory.
-If you are tight on memory, or your usage pattern does not require this, you can
-reduce these [RocksDB settings](programs-arangod-options.html#rocksdb):
+Note that it does not make sense to set to set this limit smaller than
```
jsteemann, line 189:
```suggestion
Note that it does not make sense to set this limit smaller than
```

3.11/administration-reduce-memory-footprint.md
```diff
@@ -15,123 +15,436 @@ Do not apply any of the changes suggested here before you have tested them in
...
+```
+
+The default is 1GB which essentially switches the compression off.
+If you set this to something like 100, values with at least 100 bytes
+will be compressed.
+
+This will cost a little CPU power but the system can then cache more
+data in the same amount of memory.
+
+The behaviour of the hash tables used in the caches can be adjusted with
```
jsteemann, line 410:
```suggestion
The behavior of the hash tables used internally by the caches can be adjusted with
```

3.11/administration-reduce-memory-footprint.md
```diff
@@ -15,123 +15,436 @@ Do not apply any of the changes suggested here before you have tested them in
...
+files and throw out old, no-longer used data. These compaction
+operations also need considerable amounts of RAM. If this is limited too
+much (by reducing the number of concurrent compaction operations), then
+a compaction debt can occur, which also results in write stalls or
+stops.
+
+The other subsystems outside the storage engine also need RAM. If an
+ArangoDB server queues up too many requests (for example, if more
+requests arrive per time than can be executed), then the bodies of these
+requests are cached in RAM.
```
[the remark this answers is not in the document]
neunhoef: Done.

3.11/administration-reduce-memory-footprint.md
```diff
@@ -15,123 +15,436 @@ Do not apply any of the changes suggested here before you have tested them in
...
+
+Furthermore, there are multiple different caches which are sitting in
+front of the storage engine, the most prominent one being the edge cache
+which helps to speed up graph traversals. The larger these caches,
+the more data can be cached and the higher the likelyhood that data
```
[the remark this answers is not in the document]
neunhoef: Done.

3.11/administration-reduce-memory-footprint.md
neunhoef: Done.

3.11/administration-reduce-memory-footprint.md
```diff
@@ -15,123 +15,436 @@ Do not apply any of the changes suggested here before you have tested them in
...
+--rocksdb.max-transaction-size
+```
+
+which is by default unlimited. This setting limits the total size of
+a single RocksDB transaction. If the table builder exceeds this size,
+the transaction is automatically aborted. Note that this cannot guard
+against **many** simultaneously uncommitted transactions.
+
+Another way to limit actual transaction size is "intermediate commits".
+This is a setting which leads to the behaviour that ArangoDB will
```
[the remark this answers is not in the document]
neunhoef: Done.

3.11/administration-reduce-memory-footprint.md
```diff
@@ -15,123 +15,436 @@ Do not apply any of the changes suggested here before you have tested them in
...
+automatically commit large transactions during their execution. This of
```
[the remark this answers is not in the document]
neunhoef: Done.

3.11/administration-reduce-memory-footprint.md
neunhoef: done.

3.12/administration-reduce-memory-footprint.md
```diff
@@ -15,123 +15,445 @@ Do not apply any of the changes suggested here before you have tested them in
...
+other write operations, and can thus be limited there. When there are
+many or large open transactions, this can sum up to a large amount of
+RAM usage.
+
+Once a transaction is committed and data is supposed to written out to
+level 0 SST files, there is RAM usage for table builders.
+
+As mentioned above in the block cache section, RAM usage for table
+builders can be accounted for with the block cache, if
+`--rocksdb.reserve-table-builder-memory` is set to `true`.
```
jsteemann, line 310: I don't think this should be mentioned here like this.
Transaction are built up entirely in memory before they are committed, and then the data from memory is written to the WAL and memtables. The memtables will only be flushed if their size or number exceeds whatever configured value there is. And only upon memtable flushing we will need a table builder.
If a transaction does not commit or its commit does not trigger a flush, there won't be any table builders involved.
Table builders will be used during flushes and compaction though.

neunhoef: Removed. The option is mentioned in the block cache section, so we should be good.

nerpaula (approved):

PR #1459, opened 2023-08-30
Don't use 1., 2. et cetera in headlines

Simran-B:
There is currently a mix of `datacenter` (majority) vs. `data center` but the Google style guide suggests `data center`.
Shall we address this in this PR?
- When referring to ArangoSync, `Datacenter-to-Datacenter Replication` -> `DataCenter-to-DataCenter-Replication`
- Otherwise `datacenter` -> `data center`


nerpaula:
The spelling of `datacenter` will be addressed in https://arangodb.atlassian.net/browse/DOC-537

PR #1463, opened 2023-09-01
Fix and improve arangosh examples

Simran-B:
Should be merged at the same time as https://github.com/arangodb/docs-hugo/pull/172
because of a fix for an arangosh user example.

Requires example re-generation for 3.10+


Simran-B:
https://jenkins.arangodb.biz/view/Documentation/job/arangodb-ANY-examples/3597/

PR #1468, opened 2023-09-06
Use the correct heading level for allowRetry

goedderz:
All other options use `####`, only `allowRetry` used `###`. This changes the latter to the former.


Simran-B (approved):
Good catch!

PR #1475, opened 2023-12-01
OAS-6641 | remove email_verified field from saml + scim requirement

lakhansamani:


Simran-B:
https://github.com/arangodb/docs-hugo/pull/367

